PR #17, opened 2025-02-07
Update ForkingPickler import

maanug-nv:
Update import of `ForkingPickler` to obtain from `multiprocessing` instead of `torch.multiprocessing` after this [change](https://github.com/pytorch/pytorch/commit/abd16a8c642733bd88e0bb1634b04972a8cfd688)


srogawski-nvidia (approved):
lgtm

PR #28, opened 2025-04-15
Fix thread safety in buffer access

rhewett-nv:


rhewett-nv:
We've discussed internally and decided that this update is not necessary at this time because,
* There is only one writer, even though there are multiple readers
* The timestamp reporting/use is robust to slight staleness in the timestamp and count because the injection and timestamp comparison time-windows are sufficiently different in scale

PR #29, opened 2025-04-16
Use MSC in the checkpointing

shunjiad:
Integrate the [multi-storage-client](https://github.com/NVIDIA/multi-storage-client) into the checkpoint module to support reading and writing checkpoints on object storage.

```
FileSystemWriterAsync(path="msc://test-bucket-s3/checkpoints/", use_msc=True)
```

### DIFF
--- a/src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
+++ b/src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
@@ -17,6 +17,7 @@
 import dataclasses
 import logging
 import os
+import pickle
 import queue
 from functools import partial
 from heapq import heappop, heappush
@@ -31,6 +32,7 @@ import torch
 from torch import multiprocessing as mp
 from torch.distributed.checkpoint import FileSystemWriter
 from torch.distributed.checkpoint.filesystem import DEFAULT_SUFFIX, _StoragePrefix, _write_item
+from torch.distributed.checkpoint.metadata import Metadata
 
 try:
     from torch.distributed.checkpoint.filesystem import _StorageWriterTransforms
@@ -77,8 +79,13 @@ class FileSystemWriterAsync(FileSystemWriter):
     (intermediate state is stored as writer attributes).
     """
 
-    def __init__(self, *args, separation_hint: Optional[str] = None, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(
+        self, path: Union[str, os.PathLike], *args, separation_hint: Optional[str] = None, **kwargs
+    ):
+        self.checkpoint_dir = path
+        self.use_msc = kwargs.pop("use_msc", False)
+
+        super().__init__(path, *args, **kwargs)
         if not self.single_file_per_rank:
             raise NotImplementedError(
                 'single_file_per_rank flag not supported for FileSystemWriterAsync'
@@ -159,7 +166,11 @@ class FileSystemWriterAsync(FileSystemWriter):
                 if len(bytes_data) > 0 or len(tensor_data) > 0:
                     file_name = gen_file(prefix=group_name)
                     self.write_buckets.append(
-                        (self.path / file_name, file_name, (bytes_data, tensor_data))
+                        (
+                            os.path.join(self.checkpoint_dir, file_name),
+                            file_name,
+                            (bytes_data, tensor_data),
+                        )
                     )
 
         # Check if there is anything to write on this rank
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -40,6 +40,7 @@ pyyaml = "*"
 pynvml = ">=12.0.0"
 nvidia-ml-py = ">=12.570.86"
 defusedxml = "*"
+multi-storage-client = ">=0.18.0"
```
apaithankar: Any reason this needs to be a hard dependency on NVRx?
Can we make it optional and raise an exception when user actually uses that backend (if not installed)?

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -26,11 +27,13 @@
 from time import time
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
+import multistorageclient as msc
```
apaithankar: This will change if we make it an optional dependency.

tests/checkpointing/unit/test_async_writer.py
```diff
@@ -68,17 +76,27 @@ def sync_save_checkpoint(self, checkpoint_dir, state_dict, planner):
...
-            storage_reader=FileSystemReader(checkpoint_dir),
+            storage_reader=storage_reader,
             planner=DefaultLoadPlanner(),
         )
         return state_dict
 
     def test_async_is_equivalent_to_sync(self, tmp_path_dist_ckpt):
         """Verifies that async checkpointing produces the same results as sync checkpointing."""
+        self.perform_save_and_load(tmp_path_dist_ckpt, use_msc=False)
+        self.perform_save_and_load(tmp_path_dist_ckpt, use_msc=True)
```
apaithankar: Please make this a separate test `test_msc_async_is_equivalent_to_sync`

src/nvidia_resiliency_ext/checkpointing/msc/filesystem_msc.py
```diff
@@ -0,0 +1,179 @@
...
+        # group requests by file
+        per_file: Dict[str, List[ReadItem]] = {}
+        for read_item in plan.items:
+            item_md = self.storage_data[read_item.storage_index]
+            path = item_md.relative_path
+            per_file.setdefault(path, []).append(read_item)
+
+        # prefetch objects
+        if len(per_file) > 0:
+            with ThreadPoolExecutor(max_workers=len(per_file)) as executor:
```
apaithankar: Might be a good idea to limit the `max_workers` depending on what this call actually does. E.g.:
```
max_workers=min(64, len(per_file))
```

src/nvidia_resiliency_ext/checkpointing/msc/filesystem_msc.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+    Prefetch the checkpoint files
+
+    Args:
+        checkpoint_dir (str): The checkpoint directory.
+        relative_path (str): The relative path to the checkpoint file.
+    """
+    path = os.path.join(checkpoint_dir, relative_path)
+    s = time()
+    with msc.open(path, "rb") as _:
```
apaithankar: Will this cause a file to be read?

pyproject.toml
shunjiad: The package itself is very lightweight and mainly contains interfaces. By default, it does not install any object store dependencies. Do you see any concerns if we make it a hard dependency?

src/nvidia_resiliency_ext/checkpointing/msc/filesystem_msc.py
shunjiad: The code will prefetch the checkpoint file to the local cache if the cache is configured with MSC. This is the optimization when the checkpoint file is stored on object storage.

apaithankar (approved):
LGTM

sbak5:
Please pause until either one of Slawek and myself does review and final pass.

src/nvidia_resiliency_ext/checkpointing/msc/filesystem_msc.py
scravek, line 1: If I understand correctly, this file is duplicated from megatron/core/dist_checkpointing/strategies/filesystem_msc.py.
Do we have a plan for resolving this duplication going forward?

scravek, line 1: Can we clarify whether this file should live here, in MCore, or be duplicated in both places?

skierat, line 1: Can we clarify whether this file should live here, in MCore, or be duplicated in both places?

sbak5:
LGTM

sbak5 (approved):

PR #31, opened 2025-04-16
Inprocess: update description of Progress Watchdog and Known Issues section

szmigacz:


docs/source/inprocess/usage_guide.rst
```diff
@@ -10,7 +10,8 @@ functionality provided by the :py:class:`Wrapper`.
 
 Requirements
 ------------
-In-process restart functionality requires `PyPI PyTorch
+In-process restart functionality requires `NCCL
+<https://github.com/NVIDIA/nccl>`_ v2.26.2 or higher, `PyPI PyTorch
 <https://pypi.org/project/torch/>`_ v2.5.1 through v2.6.0 or `PyTorch NGC
 Container <https://catalog.ngc.nvidia.com/orgs/nvidia/containers/pytorch>`_
 versions 24.07 through 24.10. For further limitations and compatibility
```
rhewett-nv: 25.02 / PyT 2.7 is also compatible

szmigacz: PyTorch 2.7 wasn't yet released, so we can't claim compatibility. Anything post 2.6 is tagged "2.7.0a*", but it's not actual 2.7.0, it's "post 2.6".

Regarding NGC container versions: I'm not sure what to do, versions 24.11+ were excluded because of NCCL, but we're saying "NCCL 2.26.2+" is a requirement, so out of the box none of those containers will work (including 25.02 and 25.03) because the default NCCL will be too old.  So we could say "PyTorch v2.5.1+  and NCCL 2.26.2+" without referring to actual DLFW container versions, would that work?

rhewett-nv: That's a good point.  Just listing the required versions is the most accurate way to do it

szmigacz: I pushed an update. I removed container versions and requirements are PyTorch >= 2.5.1 and NCCL >= 2.26.2

PR #36, opened 2025-04-28
Fix the sync issue in `TemporalAsyncCaller`

sbak5:
Fix sync issue for TemporalAsyncCaller, which leads to incomplete checkpoints at some cases


src/nvidia_resiliency_ext/checkpointing/async_ckpt/core.py
```diff
@@ -161,7 +161,7 @@ def close(self):
         logger.info(f"AsyncCaller: {torch.distributed.get_rank()}, Destroying Async Caller")
 
     def __del__(self):
```
rhewett-nv, line 164: This should an abstractmethod

sbak5: Missed the decorator when moved manually from Mcore. Thanks.

PR #40, opened 2025-05-06
Synchronize changes in MCore to NVRx for Metadata caching and newer changes in PyT DCP routines

sbak5:
This PR includes changes made for refactoring of metadata caching in Mcore and PyT DCP routine signature changes

### DIFF
--- a/src/nvidia_resiliency_ext/checkpointing/async_ckpt/cached_metadata_filesystem_reader.py
+++ b/src/nvidia_resiliency_ext/checkpointing/async_ckpt/cached_metadata_filesystem_reader.py
@@ -1,0 +1,51 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+""" FS Reader with metadata cached support. """
+
+import os
+from typing import Union
+
+from torch.distributed.checkpoint import FileSystemReader, Metadata
+
+
+class CachedMetadataFileSystemReader(FileSystemReader):
+    """
+    Extends FileSystemReader to cache metadata for improved performance.
+
+    Attributes:
+        _cached_metadata (Metadata or None): Cached metadata from the file system.
+    """
+
+    def __init__(self, path: Union[str, os.PathLike]) -> None:
+        """
+        Initialize with file system path.
+
+        Args:
+            path (Union[str, os.PathLike]): Path to the checkpoint directory or file.
+        """
+        super().__init__(path=path)
+        self._cached_metadata = None
+
+    def read_metadata(self) -> Metadata:
+        """
+        Read metadata from file system, caching for subsequent calls.
+
+        Returns:
+            Metadata: Checkpoint metadata.
+        """
+        if self._cached_metadata is None:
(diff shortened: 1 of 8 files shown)


sbak5:
a new test is needed to verify metadata.

apaithankar:
Will the test be added in a separate pull request?

src/nvidia_resiliency_ext/checkpointing/async_ckpt/core.py
```diff
@@ -160,7 +160,6 @@ def close(self):
         """Terminate the async caller at exit of an application or some termination conditions"""
         logger.info(f"AsyncCaller: {torch.distributed.get_rank()}, Destroying Async Caller")
 
-    @abstractmethod
```
rhewett-nv: Keep this

sbak5:
> Will the tests be added in a separate pull request?

Will move two unit tests from Mcore for metadata caching and add an example for metadata caching.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/cached_metadata_filesystem_reader.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+        Initialize with file system path.
+
+        Args:
+            path (Union[str, os.PathLike]): Path to the checkpoint directory or file.
+        """
+        super().__init__(path=path)
+        self._cached_metadata = None
+
+    def read_metadata(self) -> Metadata:
```
rhewett-nv, line 42: Is there ever a usecase where we want to flush the cache and reread it?

src/nvidia_resiliency_ext/checkpointing/async_ckpt/state_dict_saver.py
```diff
@@ -50,53 +50,248 @@ def _compare_dataclasses(obj1, obj2):
...
+        # which aggregates local plans from all ranks
+        self.cached_central_plan: SavePlan = None
+        # cached outcome of `SavePlan.prepare_local_plan` describes how local state_dict is written
+        self.cached_local_plan: SavePlan = None
+        # Cached global metadata, only `coordinator` for dist-ckpt holds
+        # if central plans are consistent over iters
+        self.cached_global_metadata: Metadata = None
+        # This variable records if the ckpt structures are consistent
+        # so the following checkpoint savings reuse `cached_global_metadata`
+        self.validated_cache_reuse: bool = False
```
rhewett-nv, line 88: I don't understand the use of the word `validated` in this var and the next.  Can we make them (and the comments) more clear?

src/nvidia_resiliency_ext/checkpointing/utils.py
```diff
@@ -127,3 +117,75 @@ def wrapped(state_dict, *args, **kwargs):
             fn(state_dict, *args, **kwargs)
 
     return wrapped
+
+
+# Dict utils copied from megatron.core.dist_checkpointing.dict_utils.py
+def diff(x1: Any, x2: Any, prefix: Tuple = ()) -> Tuple[list, list, list]:
+    """Recursive diff of dicts.
```
rhewett-nv, line 125: This function works for types that are not `dict`.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/cached_metadata_filesystem_reader.py
sbak5: This is used to bypass any collective comm for metadata when the job resumes from the previous checkpoint. 
We load a checkpoint and keep it in some variable, which is used to fill out the metadata needed for the upcoming saving.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/state_dict_saver.py
sbak5: validated means the cached metadata aligns with state dict over iterations so it's safe to reuse.

sbak5: I can add more words to the comment.

src/nvidia_resiliency_ext/checkpointing/utils.py
sbak5: It's intended. It's intended to compare elements recursively over iteratable objects.

hexinw-nvidia (approved):

PR #43, opened 2025-05-08
Added doc on restriction for TE abort usage

rhewett-nv:
This clarifies the required usage of the AbortTransfomerEngine callback.

As the new doc says, if you use the transformer engine abort callback, you need to do that after the torch distributed abort callback, otherwise restarts can hang.


rhewett-nv:
As the new doc says, if you use the transformer engine abort callback, you need to do that _after_ the torch distributed abort callback.

hexinw-nvidia (approved):

PR #44, opened 2025-05-08
Enhanced InJob restart in RankMonitorServer and launcher:

hexinw-nvidia:
- Modified read_obj_from_ipc_stream to raise exceptions instead of returning None, improving error handling clarity and robustness.
- Updated RankMonitorServer to handle launcher IPC connections, including explicit connection closure on "close_worker_ipc_connection" message.
- Added launcher IPC server to RankMonitorServer to receive commands from the launcher.
- Enhanced LocalElasticAgent to send close messages to rank monitors before shutting down workers, ensuring graceful connection closure.
- Updated RankMonitorStateMachine to allow FINALIZED state transitions in FINALIzED state. RankMonitorServer can receive SIGNAL before handling worker IPC connection lost.
- Adjusted unit tests to reflect valid signal handling in FINALIZED state.
- Added proper cleanup of launcher IPC socket file in RankMonitorServer shutdown.
- Added RankMonitorServer unit test.

### DIFF
--- a/src/nvidia_resiliency_ext/fault_tolerance/launcher.py
+++ b/src/nvidia_resiliency_ext/fault_tolerance/launcher.py
@@ -12,6 +12,8 @@
 # - Changed shutdown logic
 # - security fix for watchdog_file_path
 
+import asyncio
+
 # fmt: off
 import collections
 import contextlib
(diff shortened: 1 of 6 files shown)


src/nvidia_resiliency_ext/fault_tolerance/launcher.py
```diff
@@ -544,6 +550,24 @@ def _log_watchdog_event(
     @prof
     def _stop_workers(self, worker_group: WorkerGroup) -> None:
         logger.info(f"Stopping workers... Timeout = {self._workers_stop_timeout} sec.")
+
+        # Send close message to rank monitors
+        for local_rank, rmon_proc in self._local_rank_to_rmon.items():
+            try:
+                launcher_to_rmon_socket = f"{tempfile.gettempdir()}/_ft_launcher{rmon_proc.pid}_to_rmon.socket"
```
rhewett-nv, line 557: It is usually safer to use `tempfile.mkdtemp()` to ensure that there are no inadvertent collisions in temporary files.

rhewett-nv (approved):
The change I suggest is not fatal to the PR or necessary for merge, but it is good practice if we can safely do it.

src/nvidia_resiliency_ext/fault_tolerance/launcher.py
hexinw-nvidia: Agreed. I used "rmon_proc.pid" to avoid the collisions without changing the RankMonitorServer constructor. Also server will unlink the IPC path before using it again. I think we are safe here. Thanks.

PR #64, opened 2025-05-15
Rhewett/ftl consistency

rhewett-nv:
This PR addresses consistency in the arguments to ft_launcher.

* Some arguments are marked as deprecated and replaced with consistent use of underscore and hyphen.
* The flag to ignore missing config is removed
* Tests are cleaned up, as are docs and examples
* Vacuous `param` is removed from param args


hexinw-nvidia (dismissed):

hexinw-nvidia (approved):

sbak5 (approved):

apaithankar (approved):
Massive change. Thanks for the cleanup.

PR #80, opened 2025-05-20
Rhewett/fi improvements

rhewett-nv:
Adds some timing info to fault injection and makes a way for the workload to force clear the sentinel in case it didn't get cleared for some reason.

### DIFF
--- a/src/nvidia_resiliency_ext/inprocess/tools/inject_fault.py
+++ b/src/nvidia_resiliency_ext/inprocess/tools/inject_fault.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import ctypes
+import datetime
 import enum
 import logging
 import multiprocessing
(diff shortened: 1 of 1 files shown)


hexinw-nvidia (approved):

src/nvidia_resiliency_ext/inprocess/tools/inject_fault.py
```diff
@@ -144,6 +147,13 @@ def maybe_raise_workload_exception():
         raise InjectedException
 
 
+def clear_workload_exception():
```
apaithankar, line 150: Where is this function used?

rhewett-nv: It is used in the workload.  There are cases where the sentinel doesn't clear across restarts so in order to prevent premature faulting on non-selected ranks, we need to let the workload clear the sentinel.

apaithankar (approved):
LGTM

PR #87, opened 2025-05-20
Make MSC changes in FileSystemWriterAsync compatible with PyT 2.3

shunjiad:
The `storage_meta` was only introduced in PyTorch 2.4. For PyTorch 2.3, we can ignore this attribute.

### DIFF
--- a/src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
+++ b/src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
@@ -481,7 +481,10 @@ class FileSystemWriterAsync(FileSystemWriter):
                 storage_md.update({wr.index: wr.storage_data for wr in wr_list})
 
             metadata.storage_data = storage_md
-            metadata.storage_meta = self.storage_meta()
+
+            # storage_meta was introduced since PyTorch 2.4
+            if hasattr(self, "storage_meta"):
+                metadata.storage_meta = self.storage_meta()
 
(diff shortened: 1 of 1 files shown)


src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -481,7 +481,10 @@ def finish(self, metadata: Metadata, results: List[List[WriteResult]]) -> None:
                 storage_md.update({wr.index: wr.storage_data for wr in wr_list})
 
             metadata.storage_data = storage_md
-            metadata.storage_meta = self.storage_meta()
+
+            # storage_meta was introduced since PyTorch 2.4
+            if hasattr(self, "storage_meta"):
```
apaithankar: What is impact on MSC is `storage_meta` is not available?

shunjiad: For PyTorch 2.3, it does not need the `storage_meta` attribute.

https://github.com/pytorch/pytorch/blob/v2.3.1/torch/distributed/checkpoint/filesystem.py#L528

sbak5: This field is actually extra filed to fill in. 

Can you use `inspect.signature` instead? We use this in different places of ckpt.

sbak5 (approved):

hexinw-nvidia (approved):

PR #98, opened 2025-06-11
Making `PersistentAsyncCaller` as the default async impl to prevent issues raised by `fork` 

sbak5:


sbak5:
This includes some changes which are made in MCore for MSC.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -61,9 +61,8 @@
 def _get_write_results_queue():
     global _results_queue
     if _results_queue is None:
-        ctx = mp.get_context('fork')
-        with _disable_gc():
```
rhewett-nv, line 65: Should we disable gc or not?

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -306,7 +350,7 @@ def write_preloaded_data_multiproc(
             count_queue.join()
             # At this point, all workers completed, so the queue should have exactly
             # `len(write_buckets)` items
```
rhewett-nv, line 352: This comment no longer aligns with the code

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
sbak5: I don't think it's required for `spawn` because the manager process doesn't have any CUDA stuff and it's fine to let it do GC even if it has.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
sbak5: I'll correct this comment.

sbak5:
@aartibasant Closing this PR because other PRs have already covered the changes in this PR

PR #99, opened 2025-06-17
Added monitor_process_pidfile parameter to store monitor_process PID.

hexinw-nvidia:
Allow user to store InProcess monitor_process PID. "wait_daemon.sh" can be asked to monitor on the given monitor process.


rhewett-nv:
I like the change.  I think the `wait_daemon.sh` file is missing though

hexinw-nvidia:
> I like the change. I think the `wait_daemon.sh` file is missing though

The wait_daemon.sh is in Megatron-LM repo. I agreed it's better to be in NVRx repo.

hexinw-nvidia:
Moved wait_daemon into NVRx. Here is the updated sbatch version to use the wait_daemon:

_srun <srun_options> bash -c "
            python -u ${run_cmd}
            ret=$?
            python -m nvidia_resiliency_ext.shared_utils.wait_daemon /tmp/inprocess_monitor_\${SLURM_PROCID}.pid
            exit $ret"_

rhewett-nv:
i think this is a great upgrade for transparency and usability, thanks!

rhewett-nv (approved):

sbak5 (approved):

PR #103, opened 2025-07-02
InProcess torch.cuda.empty_cache() in Finalize should apply to the right device

hexinw-nvidia:
PyTorch maintains the currently active CUDA device in thread-local storage. Each Python thread needs torch.cuda.set_device to the right device.


rhewett-nv (approved):
looks great!

src/nvidia_resiliency_ext/inprocess/finalize.py
```diff
@@ -93,11 +97,28 @@ def __init__(
 
     def __call__(self, state: FrozenState) -> FrozenState:
         rank = state.rank
+
+        # Determine device in main thread before creating the thread
+        device_id = None
+        if torch.cuda.is_available() and torch.cuda.is_initialized():
```
apaithankar: should we throw exception if this is false?

src/nvidia_resiliency_ext/inprocess/finalize.py
```diff
@@ -93,11 +97,28 @@ def __init__(
...
+            if (local_rank := os.getenv('LOCAL_RANK', None)) is not None:
+                device_id = torch.device(int(local_rank))
+            else:
+                device_id = torch.device(torch.cuda.current_device())
+
+        def wrapped_fn():
+            # Set CUDA device in the thread
+            if device_id is not None:
```
apaithankar, line 115: Accessing global variable from thread. Should this be instead passed in as a thread parameter?

src/nvidia_resiliency_ext/inprocess/finalize.py
hexinw-nvidia: I agree, it should, just that the exception in general will be swallowed quietly in the thread other than the main thread. We'd need to have a decorator around the current thread entrance to log the exception and maybe exit the whole process to raise enough attention to users to fix the SW bug.

src/nvidia_resiliency_ext/inprocess/finalize.py
hexinw-nvidia: The current approach is also fine. The device_id is global inside the __call__ scope. It is defined a couple of lines above the wrapped_fn. It is clear enough for the user to be aware of the code path.

src/nvidia_resiliency_ext/inprocess/finalize.py
rhewett-nv: This could be done at creation of the `Finalize` object, which happens before the wrapped function is called.

src/nvidia_resiliency_ext/inprocess/finalize.py
```diff
@@ -93,11 +97,28 @@ def __init__(
...
+
+        # Determine device in main thread before creating the thread
+        device_id = None
+        if torch.cuda.is_available() and torch.cuda.is_initialized():
+            if (local_rank := os.getenv('LOCAL_RANK', None)) is not None:
+                device_id = torch.device(int(local_rank))
+            else:
+                device_id = torch.device(torch.cuda.current_device())
+
+        def wrapped_fn():
```
apaithankar, line 113: Any reason to not get the device_id in the thread itself?

src/nvidia_resiliency_ext/inprocess/finalize.py
```diff
@@ -93,11 +97,28 @@ def __init__(
 
     def __call__(self, state: FrozenState) -> FrozenState:
         rank = state.rank
+
+        # Determine device in main thread before creating the thread
+        device_id = None
+        if torch.cuda.is_available() and torch.cuda.is_initialized():
+            if (local_rank := os.getenv('LOCAL_RANK', None)) is not None:
+                device_id = torch.device(int(local_rank))
+            else:
```
apaithankar: We also have `rank` from `state.rank` which could be used.

src/nvidia_resiliency_ext/inprocess/finalize.py
hexinw-nvidia: Do you mean to call "torch.cuda.current_device" from inside the thread? That is exactly the problem we are trying to fix. The CUDA device context is thread-local.

src/nvidia_resiliency_ext/inprocess/finalize.py
hexinw-nvidia: Is "state.rank" a global rank or local rank?

src/nvidia_resiliency_ext/inprocess/finalize.py
hexinw-nvidia: I fixed the code to raise an exception. PTAL.

apaithankar (approved):

apaithankar:
LGTM!

PR #108, opened 2025-07-04
make the persistent async cp worker default

aartibasant:
1. Update examples to make the persistent async CP worker default
2. Throw warnings in case the TemporalAsync Worker is used instead
3. This is to discourage the use of `fork` in the TemporalAsyncWorker which is known to have instabilities. Use the more reliable `spawn` based PersistentAsyncWorker

Co-authored-by: sbak@nvidia.com


examples/checkpointing/async_ckpt.py
```diff
@@ -23,9 +23,14 @@ def parse_args():
         help="Checkpoint directory for async checkpoints",
     )
     parser.add_argument(
-        '--persistent_queue',
-        action='store_true',
-        help="Enables a persistent version of AsyncCallsQueue.",
+        '--no_persistent_queue',
```
apaithankar, line 26: Why are we providing the option to disable persistent queue? Why have any option around this at all?

sbak5: I think Aarti has moved this option I added in PR #98. 
@aartibasant Please remove these options in `examples` so these examples run on the persistent worker by default.

aartibasant: I kept the option active due to the following assertion warning: 
https://github.com/NVIDIA/nvidia-resiliency-ext/blob/main/examples/checkpointing/local_ckpt.py#L150-L152

Once we are able to validate replication with persistent_queue, I will remove the option completely.

apaithankar: That the option can be changed to `assert(not args.replication)`.

We might have to disable replication. Isn't temporal + replication is basically replicating a corrupted checkpoint?

aartibasant: But that means that as a side effect of this PR, we end up disabling `replication` feature completely which was never an intent in this release.

apaithankar: Okay, in that case, we can K/B this and warn users about using local replicated checkpoints. Perhaps print a big warning when that combination is used.

aartibasant: I have added a warning that `TemporalWorker will be deprecated soon`
However, I do not have the context on the issue with replication and so have not changed any warnings around the replication feature.

sbak5: We can't make a proper change for local ckpt with persistent ckpt. 
Let's skip this release and make it work later in the major release.

apaithankar: That's the reason for Kb (release note) this issue. We have no choice but to fix in later release.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/core.py
```diff
@@ -485,14 +485,15 @@ class AsyncCallsQueue:
...
-    def __init__(self, persistent: bool = False):
+    def __init__(self, persistent: bool = True):
         self.async_calls: deque[_ActiveAsyncRequest] = deque([])
         self.call_idx: int = -1
         self.persistent: bool = persistent
         self.persistent_caller: AsyncCaller = None
 
     def _get_async_caller(self):
         if not self.persistent:
+            logger.warning("The TemporalAsyncCaller will be deprecated soon. ")
```
apaithankar, line 496: The comment should read: "The TemporalAsyncCaller is deprecated and will be removed soon".

apaithankar: Or is it not being deprecated in this release?

aartibasant: It is not deprecated in this release due to the `replication` feature that relies on it.

rhewett-nv (approved):

apaithankar (approved):
LGTM!

PR #109, opened 2025-07-04
fix sync CP to preload tensors

aartibasant:
The sync Save utility must preload tensors before triggering the persist checkpoint function.

Co-authored-by:  sbak@nvidia.com


apaithankar:
Is this a bugfix? Also, is there a JIRA/nvbug associated with this?

PR #111, opened 2025-07-07
fix sync CP to preload tensors

aartibasant:
The sync Save utility must preload tensors before triggering the persist checkpoint function.

Co-authored-by: sbak@nvidia.com


src/nvidia_resiliency_ext/checkpointing/async_ckpt/core.py
```diff
@@ -78,11 +78,18 @@ def execute_sync(self) -> None:
...
+        # preload tensors.
+        async_fn_args = list(self.async_fn_args)
+        if self.preload_fn:
+            # the 2nd arg is state dict
+            async_fn_args[1] = self.preload_fn()
+        # persist the state
         if self.async_fn is not None:
-            self.async_fn(*self.async_fn_args)
+            self.async_fn(*async_fn_args, **self.async_fn_kwargs)
+        # This utility implements a sync cp save. Hence the barrier.
```
sbak5, line 89: Can you add `del async_fn_args`?

One of the reason is we stage out tensors in `async_fn_args` to a local variable and they're dereferenced. Until the trainer calls `gc.collect()`, this dereferenced object may not be reclaimed to PyT internal memory pool. 

to minimize a random jitter incurred by GC, MLM runs with GC turned off.

sbak5 (changes requested):

src/nvidia_resiliency_ext/checkpointing/async_ckpt/core.py
aartibasant: @sbak5 : the `del` statement will only dereference the object `async_fn_args ` which will happen anyways after the finalize call. The dereferenced object memory will be reclaimed only when GC is triggered. What symptom were you seeing without this `del` statement? I did some extensive testing and did not see the need for `del` statement.

sbak5 (approved):

src/nvidia_resiliency_ext/checkpointing/async_ckpt/core.py
aartibasant: Marking as resolved based on chat

apaithankar (approved):
LGTM

PR #115, opened 2025-07-10
[Straggler det.] Check for invalid timestamps

jbieniusiewi:
Recently, some instances were observed where timestamps reported by CUPTI had a value of 0.
This can break score computation logic, for example by causing division by zero errors.
This MR filters out invalid timestamps.


src/nvidia_resiliency_ext/attribution/straggler/cupti_src/CuptiProfiler.cpp
```diff
@@ -178,7 +178,11 @@ void CuptiProfiler::CUPTIAPI bufferCompleted(CUcontext ctx, uint32_t streamId, u
         status = cuptiActivityGetNextRecord(buffer, validSize, &record);
         if (status == CUPTI_SUCCESS && record->kind == CUPTI_ACTIVITY_KIND_CONCURRENT_KERNEL) {
             const auto *kernel = (const CUpti_ActivityKernel4 *)record;
-            
+
+            if (kernel->start == 0 || kernel->end == 0) {
```
hexinw-nvidia, line 182: Does this mean a bug in the Cupti? Is it also possible to have "kernel->end < kernel->start"? Thanks.

jbieniusiewi: According to the CUPTI docs:

> A value of 0 for both the start and end timestamps indicates that timestamp information could not be collected for the kernel.

https://docs.nvidia.com/cupti/api/structCUpti__ActivityKernel.html#_CPPv4N20CUpti_ActivityKernel5startE

sbak5 (approved):

PR #116, opened 2025-07-10
local checkpoint: update async usage to persistence=False

diegs:
LocalCheckpointManager AsyncRequests are incompatible with the persistent queue because some of their methods are not pickleable.

AsyncCallsQueue lifecycle is not owned by LocalCheckpointManager, so the only thing we can do to handle this is update the example and documentation.

### DIFF
--- a/examples/checkpointing/local_ckpt.py
+++ b/examples/checkpointing/local_ckpt.py
@@ -42,16 +42,6 @@ def parse_args():
         "Needs to be enabled on all ranks.",
     )
     parser.add_argument(
-        '--no_persistent_queue',
-        action='store_false',
-        default=True,
-        dest='persistent_queue',
-        help=(
-            "Disables a persistent version of AsyncCallsQueue. "
-            "Effective only when --async_save is set."
-        ),
-    )
-    parser.add_argument(
         '--replication_jump',
         default=4,
         type=int,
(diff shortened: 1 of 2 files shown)


examples/checkpointing/local_ckpt.py
```diff
@@ -162,7 +146,8 @@ def main():
 
     # Instantiate checkpointing classess needed for local checkpointing
     ckpt_manager = create_checkpoint_manager(args)
-    async_queue = AsyncCallsQueue(persistent=args.persistent_queue) if args.async_save else None
+    # Persistent queue is incompatible with local checkpointing because some routines are not pickleable.
+    async_queue = AsyncCallsQueue(persistent=False) if args.async_save else None
```
sbak5, line 150: Can you add a warning message here?

diegs: What would be the warning message? Just alerting the person running the example to the same thing I wrote in the comment?

rhewett-nv: If user-provided options are incompatible it should be an error, not a warning, and definitely not a silent.

diegs: @rhewett-nv if you look at the delta, I removed the persistence option completely based on @sbak5 's recommendation, and hardcoded persistence to false on this line. So there is no longer a possibility of an incompatible option.

Regarding library usage, as I wrote in the PR description, AsyncCallsQueue and LocalCheckpointManager are independent classes, so there is no way to enforce their correct co-usage in code.

PR #117, opened 2025-07-10
local checkpoint: update async usage to persistence=False

diegs:
LocalCheckpointManager AsyncRequests are incompatible with the persistent queue because some of their methods are not pickleable.

AsyncCallsQueue lifecycle is not owned by LocalCheckpointManager, so the only thing we can do to handle this is update the example and documentation.


diegs:
Cherry-pick of #116 into 0.4.0 branch

PR #118, opened 2025-07-13
Added in-process wrapper restart latency

namitdhameja:
Tracks min/max restart latency of the in-process wrapper

### DIFF
--- a/src/nvidia_resiliency_ext/inprocess/wrap.py
+++ b/src/nvidia_resiliency_ext/inprocess/wrap.py
@@ -401,6 +401,7 @@ class CallWrapper:
         wrapper = self.wrapper
         progress_watchdog = self.progress_watchdog
 
+        exception_recvd_time = None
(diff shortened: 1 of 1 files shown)


src/nvidia_resiliency_ext/inprocess/wrap.py
```diff
@@ -463,6 +464,42 @@ def __call__(self, fn, args, kwargs):
...
 
                             if state.mode == Mode.ACTIVE:
+                                if exception_recvd_time is not None:
+                                    # Store local trigger time in the distributed store
+                                    store.set(
+                                        f"exception_recvd_time_{state.rank}",
+                                        str(exception_recvd_time),
+                                    )
+                                    # Barrier to ensure all ranks have set their trigger time
+                                    store.completion_barrier(
```
rhewett-nv, line 474: The barriers defined in `Store` can only be used once.  To add a barrier here we'll need to add another to the `StoreMixin`.

PR #119, opened 2025-07-15
Inprocess doc clarification

rhewett-nv:
Update inprocess docs to clarify:
* Where RankAssignment should be relative to Nested Restarts in Initialize
* Why the "optimal" example is safe, re: process-group independence.

### DIFF
--- a/docs/source/inprocess/usage_guide.rst
+++ b/docs/source/inprocess/usage_guide.rst
@@ -60,7 +60,10 @@ Requirements for the wrapped function
   ext/blob/main/examples/fault_tolerance/run_inprocess_injob_example.sh>`_ example for the recommended
   default values (for example, --monitor-interval=5).
 
-- it's heavily recommended for the wrapped function to load the state affected
+    - Any objects whose lifetime crosses the restart boundary *must* be process-group independent _or_
+      they are responsible for re-aligning themselves to the new process groups within the wrapped function.
+
+- It is heavily recommended for the wrapped function to load the state affected
   by distributed collectives from a checkpoint on every restart (e.g. load
   weights of a model); outputs of distributed collectives are likely to become
   corrupted or invalid if a fault happened while a collective was in-flight and
--- a/docs/source/inprocess/usage_guide.rst
+++ b/docs/source/inprocess/usage_guide.rst
@@ -313,7 +313,10 @@ workers drops below a specified threshold.
 Multiple initializers could be composed with :py:class:`nvidia_resiliency_ext.inprocess.Compose`.
 The composition order follows mathematical composition. Therefore, the last listed function is called first.
 Consequently, when using nested restarters, the :py:class:`nvidia_resiliency_ext.inprocess.nested_restarter.NestedRestarterHandlingCompleted`
-should be listed first, as handling a restart is not complete until the end of the `Initialize`.
+should be carefully placed as this callback indicates completion of the restart.
+Subsequent callbacks (e.g., those listed before the nested restarter callback) logically take placed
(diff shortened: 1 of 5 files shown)


examples/inprocess/optimal_example.py
```diff
@@ -236,8 +236,9 @@ def main():
     else:
         raise RuntimeError
 
-    # All objects created in ``main()`` are constructed only once, and reused
-    # for all restart iterations.
+    # In this example, all objects created in ``main()`` are constructed only once and reused
+    # for all restart iterations. *This* model and optimizer are process-group independent,
+    # so they can safely be created outside of the wrapped function and reused.
```
ShreyaGupta08: I think we should either
- make it explicit that if a user is using tools that create process groups, then they should wrap the model definition etc. in the wrapper (i.e. adding something like "if you use libraries or enable functionalities that create process groups, you should wrap model definition and declarations in the wrapper. More guidelines: Docs" (while pointing to the docs changes above), or
- wrap everything in the wrapper in this example itself

docs/source/inprocess/usage_guide.rst
```diff
@@ -60,7 +60,10 @@ Requirements for the wrapped function
   ext/blob/main/examples/fault_tolerance/run_inprocess_injob_example.sh>`_ example for the recommended
   default values (for example, --monitor-interval=5).
 
-- it's heavily recommended for the wrapped function to load the state affected
+    - Any objects whose lifetime crosses the restart boundary *must* be process-group independent _or_
+      they are responsible for re-aligning themselves to the new process groups within the wrapped function.
```
ShreyaGupta08: The structure of this feels complex; can we make the language simple and accessible?

Something on the lines of "If the program initializes process groups, the user is responsible for either wrapping the variables leading to the process group declaration in the wrapped function or cleanup and re-initialize their process groups themselves" (or similar variation thereof)

ShreyaGupta08 (changes requested):

docs/source/inprocess/usage_guide.rst
```diff
@@ -60,7 +60,11 @@ Requirements for the wrapped function
   ext/blob/main/examples/fault_tolerance/run_inprocess_injob_example.sh>`_ example for the recommended
   default values (for example, --monitor-interval=5).
 
-- it's heavily recommended for the wrapped function to load the state affected
+    - Any objects whose lifetime crosses the restart boundary *must* be process-group independent _or_
+      the workload must re-aligning the object, inside the wrapped function, to the new process groups
```
apaithankar: s/re-aligning/re-align

rhewett-nv: fixed, ty

examples/inprocess/optimal_example.py
ShreyaGupta08: How is this? 
In this example, all objects created in ``main()`` are constructed only once and reused for all restart iterations. if you use libraries or enable functionalities that create process groups, you should wrap model definition and declarations in the wrapper. *This* model and optimizer are process-group independent, so they can safely be created outside of the wrapped function and reused. See the usage guide for more details.

apaithankar (approved):

apaithankar (approved):

docs/source/inprocess/usage_guide.rst
apaithankar: We discussed this offline.

examples/inprocess/optimal_example.py
apaithankar: We discussed this offline

PR #122, opened 2025-07-17
Add FR trace collection at abort when FR and env vars are set

sbak5:
Adding an extra step in `AbortTorchDistributed` when FR variables are set
(TORCH_NCCL_DUMP_ON_TIMEOUT', 'TORCH_NCCL_ENABLE_MONITORING)

Then, each rank will dump its FR trace at inprocess abort phase to a directory set by `NVRX_FR_TRACE_PATH`


src/nvidia_resiliency_ext/attribution/trace_analyzer/trace_collector.py
```diff
@@ -0,0 +1,126 @@
...
+            local_rank = self.rank % torch.cuda.device_count()
+            health_check_results = TorchFRTraceCollector.get_health_check_results(local_rank)
+            dumped_dict['health_check_results'] = {
+                'gpu': health_check_results['gpu_health_check'],
+                'nic': health_check_results['nic_health_check'],
+            }
+            if self.json:
+                json.dump(dumped_dict, f, indent=4)
+            else:
+                pickle.dump(dumped_dict, f)
```
rhewett-nv, line 101: Is there any specific reason to maintain support for pickle?  Why not JSON only?

src/nvidia_resiliency_ext/attribution/trace_analyzer/trace_collector.py
```diff
@@ -0,0 +1,126 @@
...
+    Derived classes must implement `collect()` to gather trace data at runtime.
+    """
+
+    def __init__(
+        self,
+        path: str,
+    ):
+        self.path = path
+
+    def collect(self):
```
rhewett-nv, line 33: @abstractmethod

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -302,6 +302,70 @@ def _perform_health_check(self) -> bool:
...
+        """
+        Core method to perform GPU health check. Used by both sync and async checks.
+
+        Checks the recovery action needed for each GPU and ensures all GPUs are healthy.
+
+        Returns:
+            bool: True if all GPUs are healthy (no recovery action needed), False otherwise.
+        """
+        try:
+            self.pynvml.nvmlInit()
```
rhewett-nv, line 316: Is it possible to init once and maintain the context associated with it?  I worry that this can thrash the GPU memory depending on how often it is called.

src/nvidia_resiliency_ext/inprocess/abort.py
```diff
@@ -97,8 +100,27 @@ def shutdown_process_group_backend(group, device):
                 else:
                     backend.abort()
 
+    @staticmethod
+    def collect_fr_trace():
+        def _check_fr_env():
+            check_env_variables = ['TORCH_NCCL_DUMP_ON_TIMEOUT', 'TORCH_NCCL_ENABLE_MONITORING']
+            for env_var in check_env_variables:
+                if os.environ.get(env_var, '0') == '0':
+                    return False
```
rhewett-nv, line 146: Rather than silent failure, a one-time, single-rank info log will be informative for end users.

rhewett-nv (changes requested):

src/nvidia_resiliency_ext/attribution/trace_analyzer/trace_collector.py
sbak5: Json is actually optional and more bloated than binary. The original returned dict from PyT is a pickled dictionary. So, the default is pickled dictionary and json is for debugging to see each individual trace.

src/nvidia_resiliency_ext/shared_utils/health_check.py
sbak5: This applies to the original NVML health check as well. Maybe it needs a separate PR to fix this GPU health check with a global context and destruction of the global context.

src/nvidia_resiliency_ext/inprocess/abort.py
sbak5: Made a rank 0 print the warning once per job using a class variable.

src/nvidia_resiliency_ext/shared_utils/health_check.py
sbak5: I'll make a separate MR to fix this issue.

src/nvidia_resiliency_ext/attribution/trace_analyzer/trace_collector.py
```diff
@@ -0,0 +1,129 @@
+import json
+import logging
+import os
+import pickle
```
rhewett-nv: Please add this block for the pickle import.
```
# Issue: [B403:blacklist] Consider possible security implications associated with pickle module.
# Severity: Low   Confidence: High
# CWE: CWE-502 (https://cwe.mitre.org/data/definitions/502.html)
# More Info: https://bandit.readthedocs.io/en/1.8.3/blacklists/blacklist_imports.html#b403-import-pickle
import pickle  # nosec
```

src/nvidia_resiliency_ext/inprocess/abort.py
```diff
@@ -97,8 +103,40 @@ def shutdown_process_group_backend(group, device):
...
+                    log = logging.getLogger(__name__)
+                    log.info(
+                        f"Environment variable {env_var} is set to {env_value}"
+                        f", FR trace collection is disabled"
+                    )
+                    cls._logging_printed = True
+                    return False
+            return True
+
+        if _check_fr_env() is True:
```
rhewett-nv, line 149: The idiomatic way to do this in `inprocess` is to create a separate abort callback.  This way, workloads can choose to incorporate the callback or not and enable pathing for that using their CLI arguments.

If there is a good reason to have it directly as part of `AbortTorchDistributed` I am OK to keep it here.  However, I am also wary of having the on/off logic determined by the presence of the path env variable.

sbak5: The ordering should be maintained between this step and destroying process groups. 
dump FR is not available after `AbortTorchDistributed` is called.

rhewett-nv: OK, that is sensible.  Can you expose a way to enable/disable or set the paths based on the constructor for the `AbortTorchDistributed` callback then?

sbak5: Added an argument in the constructor.

rhewett-nv: I don't see the change, did you push it?

sbak5: Pushed again.

rhewett-nv (approved):

hexinw-nvidia (approved):

PR #126, opened 2025-07-18
Define a NVRx logger to address Large-Scale LLM training

hexinw-nvidia:
This module provides a simple and efficient log manager that supports both
regular logging and distributed logging for large-scale training with thousands
of GPUs. The design automatically adapts based on the environment configuration.

Key Design Principles:
- Environment-driven behavior: NVRX_DIST_LOG_DIR controls distributed vs regular logging
- Per-node aggregation: When distributed logging is enabled, local rank 0 aggregates logs
- Dynamic rank detection: Automatically reads rank info from environment variables
- Scalable: Works with 3K+ GPUs without overwhelming logging infrastructure
- Fork-safe: All ranks use file-based messaging to ensure child processes can log
- Subprocess-safe: Supports force_reset=True for fresh logger setup in subprocesses


src/nvidia_resiliency_ext/shared_utils/logger.py
```diff
@@ -0,0 +1,584 @@
...
+        return f"[{timestamp_str}] [{level_name}] [{self.hostname}:{self.rank}({self.local_rank})] {self.message}"
+
+
+class LogManager:
+    """
+    Log manager for large-scale LLM training.
+
+    Supports both regular logging and distributed logging. When distributed logging
+    is enabled (NVRX_DIST_LOG_DIR is set), each node logs independently to avoid
+    overwhelming centralized logging systems. Local rank 0 acts as the node aggregator,
```
rhewett-nv: Let's make this customizable, and if we can default to local rank P-1.  Rank 0 has so much responsibility we should avoid further overload.

src/nvidia_resiliency_ext/shared_utils/logger.py
```diff
@@ -0,0 +1,584 @@
...
+        # Use NVRX_LOG_DEBUG environment variable to determine log level
+        debug_enabled = os.environ.get("NVRX_LOG_DEBUG", "").lower() in ("1", "true", "yes", "on")
+        self.log_level = logging.DEBUG if debug_enabled else logging.INFO
+
+        # Determine if distributed logging is enabled
+        self._log_dir = log_dir or os.environ.get("NVRX_DIST_LOG_DIR", None)
+        self.distributed_logging_enabled = self._log_dir is not None
+
+        # Use shared temporary directory for pending messages (avoid Lustre for temp files)
+        # Use /tmp with node_id to ensure all ranks on the same node use the same directory
```
rhewett-nv: 1. we should use tmpfile or something similar to prevent any collisions across job starts (not every run will clear /tmp)
2. Are you sure that each container on a node will mount the same `/tmp`?

src/nvidia_resiliency_ext/shared_utils/logger.py
```diff
@@ -0,0 +1,584 @@
...
+        # Create logger
+        self._logger = self._setup_logger()
+
+        # Start aggregator if distributed logging is enabled and this is the aggregator
+        if self.distributed_logging_enabled and self._is_aggregator:
+            self._start_aggregator()
+
+    @property
+    def rank(self) -> int:
+        """Get the global rank."""
```
rhewett-nv: In NVRx we now have multiple "global" ranks.  We should emphasize that this is the global rank at the job orchestration (Slurm, etc) level, not at the workload or restarter level.

src/nvidia_resiliency_ext/shared_utils/logger.py
```diff
@@ -0,0 +1,584 @@
...
+
+            # Use dynamic formatter with static hostname and dynamic rank info
+            formatter = DynamicLogFormatter(
+                fmt=f"%(asctime)s [%(levelname)s] [{self._hostname}:%(rank)s(%(local_rank)s)] %(filename)s:%(lineno)d %(message)s"
+            )
+
+            handler.setFormatter(formatter)
+            logger.addHandler(handler)
+        else:
+            # Simple logging to stderr or file
```
rhewett-nv: s/file/stdout

src/nvidia_resiliency_ext/shared_utils/logger.py
```diff
@@ -0,0 +1,584 @@
...
+            if os.environ.get("NVRX_LOG_TO_STDOUT", "").lower() in ("1", "true", "yes", "on"):
+                handler = logging.StreamHandler(sys.stdout)
+            else:
+                handler = logging.StreamHandler(sys.stderr)
+
+            handler.setLevel(self.log_level)
```
rhewett-nv: This block is repeated in the if, can move it after the branch once `handler` is set

src/nvidia_resiliency_ext/shared_utils/logger.py
```diff
@@ -0,0 +1,584 @@
...
+                            if line:
+                                parts = line.split('\t')
+                                if len(parts) >= 6:
+                                    level, message, rank, local_rank, hostname, timestamp = parts[
+                                        :6
+                                    ]
+
+                                    log_msg = LogMessage(
+                                        level=int(level),
+                                        message=message,
```
rhewett-nv: Message contents can have `\t`.  You may need a more explicit separator, or just have the log message written as a JSONL blob.

src/nvidia_resiliency_ext/shared_utils/logger.py
```diff
@@ -0,0 +1,584 @@
...
+            # Look for message files from all ranks (including this aggregator rank)
+            for filename in os.listdir(msg_dir):
+                if filename.startswith('rank_') and filename.endswith('.msg'):
+                    msg_file = os.path.join(msg_dir, filename)
+                    try:
+                        # Get current file size
+                        try:
+                            file_size = os.path.getsize(msg_file)
+                        except (IOError, OSError):
+                            # File doesn't exist or can't be accessed, skip
```
rhewett-nv: If it doesn't exist, that means either the mounts are not visible across processes or that something went wrong.

This should at minimum log a warning to the job-level stderr.

src/nvidia_resiliency_ext/shared_utils/logger.py
```diff
@@ -0,0 +1,584 @@
...
+                            continue
+
+                        # Get the last known position for this file
+                        last_position = self._file_positions.get(msg_file, 0)
+
+                        # If file hasn't grown, skip it
+                        if file_size <= last_position:
```
rhewett-nv: What stops the log file from growing unboundedly during a long run?

Can we make use of logrotate or something similar to avoid issues?

src/nvidia_resiliency_ext/shared_utils/logger.py
```diff
@@ -0,0 +1,584 @@
...
+        # Clean up temporary directory (only if this is the last rank on the node)
+        try:
+            import shutil
+
+            if hasattr(self, '_temp_dir') and os.path.exists(self._temp_dir):
+                # Only clean up if this is the last rank (rank 0) to avoid conflicts
+                if self._is_aggregator:
+                    shutil.rmtree(self._temp_dir)
+        except (OSError, IOError):
+            # Ignore cleanup errors
```
rhewett-nv: This will make debugging difficult for this feature

src/nvidia_resiliency_ext/shared_utils/logger.py
hexinw-nvidia: Log aggregator is now a separate service from any rank.

src/nvidia_resiliency_ext/shared_utils/logger.py
hexinw-nvidia: With latest design of aggregator running as a separate service, I think it is less likely to go wrong in the log aggregation. But if needed, we can certainly add a debug flag not to delete the temporary files on exit in future.

src/nvidia_resiliency_ext/shared_utils/logger.py
hexinw-nvidia: We now printed both workload rank info and infra rank info to gain more visibility into the system.
The log format string is currently looking like the following:

`1: 2025-07-21 19:21:19,408 [INFO] [cw-dfw-h100-001-297-012] [workload:8(0) infra:1(0)] rank_monitor_server.py:171 Enable NIC health monitoring.
`

src/nvidia_resiliency_ext/shared_utils/logger.py
hexinw-nvidia: LogMessage now contains just the message itself.

src/nvidia_resiliency_ext/shared_utils/logger.py
hexinw-nvidia: Change all the exception handling in the logger to output an error rather than silently suppressing.

src/nvidia_resiliency_ext/shared_utils/logger.py
hexinw-nvidia: Added log rotate function.

.gitignore
```diff
@@ -4,3 +4,5 @@ dist
 __pycache__
 cupti_module.*.so
 .pytest_cache
+tests/shared_utils/logs
+tests/shared_utils/tmp
```
hexinw-nvidia: Do we want to have generated files under "tests/shared_utils"? Or we can point them to some temp directory outside of the source directory?

README.md
```diff
@@ -11,25 +11,26 @@ For detailed documentation and usage information about each component, please re
 
 ## Core Components and Capabilities
 
-- **[Fault Tolerance](https://github.com/NVIDIA/nvidia-resiliency-ext/blob/main/docs/source/fault_tolerance/index.rst)**
+- **[Fault Tolerance](https://nvidia.github.io/nvidia-resiliency-ext/fault_tolerance/index.html)**
```
hexinw-nvidia: These changes are probably due to out-of-sync to the main branch?

src/nvidia_resiliency_ext/inprocess/wrap.py
```diff
@@ -268,8 +269,8 @@ def __init__(self, wrapper: Wrapper):
         self.state = None
 
         try:
-            utils.Logging.initialize()
-            log = logging.getLogger(__name__)
+            setup_logger()
+            log = logging.getLogger("nvrx")
```
hexinw-nvidia: There are two OS processes in the InProcess solutions.
One is the wrapper which is the trainer process. The other is the monitor process.
We need to setup_logger() in the monitor process, too. Need to verify that monitor_process logging are now coming to the NVRx logger.

hexinw-nvidia (commented):
Namit and I discussed over the design.
There are three high level actions.
1) We can move the aggregator function out of log_manager.py. The log_manager.py will be just NVRx logger functions.
2) There will be multiple writers to a single temp per-rank log file (the training process and the monitor process). We need to look into how to do the log rotation correctly. But log rotation is not a P0 task. We can target this after basic logger change is in.
3) We should make the design flexible that all ranks (on a same node) can write to the per-node local file directly. The log aggregator would become a simple rsync or tail service to stream the local file to LFS, just to reduce the pressure on Lustre performance in scale.

examples/inprocess/basic_example.py
```diff
@@ -113,6 +113,7 @@ def parse_args():
         default=logging.INFO,
         help='logging level',
     )
+    parser.add_argument("--temp-dir", help="Directory for temporary files")
```
hexinw-nvidia: User can export NVRX_LOG_TEMP_DIR and NVRX_LOG_DEBUG directly in their sbatch script. There is no need to add one more layer of arguments. Right?

hexinw-nvidia: Also, we should not touch application logging setting. We just need to make sure NVRx logger is independent from the application logging, and can be used inside NVRx modules.

examples/inprocess/optimal_example.py
```diff
@@ -44,6 +44,7 @@
 import torch
 
 import nvidia_resiliency_ext.inprocess as inprocess
+from nvidia_resiliency_ext.shared_utils.log_manager import setup_logger
```
hexinw-nvidia: Do not force application to use NVRx logger. 
Changes are not needed.

src/nvidia_resiliency_ext/shared_utils/log_aggregator.py
```diff
@@ -0,0 +1,104 @@
...
+from nvidia_resiliency_ext.shared_utils.log_distributed import NodeLogAggregator
+import nvidia_resiliency_ext.shared_utils.log_manager as LogMgr
+
+
+def main():
+    """Main function for running the log aggregator as a separate service."""
+    parser = argparse.ArgumentParser(description="NVRx Log Aggregator Service")
+    parser.add_argument("--log-dir", help="Directory for log files")
+    parser.add_argument("--temp-dir", default='/tmp', help="Directory for temporary files)")
+    parser.add_argument("--en_chrono_ord", help="Enable Chronological Ordering")
```
hexinw-nvidia: Use "--en-chrono-ord"

src/nvidia_resiliency_ext/shared_utils/log_manager.py
```diff
@@ -0,0 +1,383 @@
...
+- Dual mode operation: Regular logging (stderr/stdout) or distributed logging (file aggregation)
+- Per-node log files: When distributed logging is enabled (e.g., node_hostname.log)
+- Automatic rank and node identification in log messages
+- Thread-safe logging with proper synchronization
+- Environment variable configuration for easy deployment
+- Fork-safe design with file-based message passing for all ranks
+- Separate aggregator service: Can run independently of training processes
+- Configurable temp directory: Customizable location for pending message files
+
+Environment Variables:
```
rhewett-nv, line 43: `NVRX_DIST_LOG_DIR` is not documented here in the env variables.

src/nvidia_resiliency_ext/shared_utils/log_manager.py
```diff
@@ -0,0 +1,383 @@
...
+    
+Note: File rotation is designed to be safe for the aggregator service. When files are rotated,
+the aggregator will automatically read from both current and backup files to ensure no messages are lost.
+
+Usage:
+    # In main script (launcher.py)
+    from nvidia_resiliency_ext.shared_utils.log_manager import setup_logger
+    logger = setup_logger()  # Call once at startup
+    
+    # In other modules
```
rhewett-nv, line 58: Missing `from nvidia_resiliency_ext.shared_utils.log_manager import LogConfig`

src/nvidia_resiliency_ext/shared_utils/log_manager.py
```diff
@@ -0,0 +1,383 @@
...
+    @classmethod
+    def get_temp_dir(cls, temp_dir=None):
+        # Use configurable temporary directory for pending messages
+        return temp_dir or os.environ.get("NVRX_LOG_TEMP_DIR")
+
+    @classmethod
+    def get_max_file_size_kb(cls, file_size_kb=None) -> int:
+        if file_size_kb is None:
+            file_size_kb = int(os.environ.get("NVRX_LOG_MAX_FILE_SIZE_KB", "10240"))
+        return file_size_kb * 1024  # Convert KB to bytes
```
rhewett-nv, line 128: Is the intended output in bytes or kb?  The method name should drop `_kb`

src/nvidia_resiliency_ext/shared_utils/log_manager.py
```diff
@@ -0,0 +1,383 @@
...
+
+    @classmethod
+    def get_max_backup_files(cls) -> int:
+        return int(
+            os.environ.get("NVRX_LOG_MAX_BACKUP_FILES", "4")
+        )  # Keep default 5 backup files per rank
```
rhewett-nv: is it 4 or is it 5?

src/nvidia_resiliency_ext/shared_utils/log_manager.py
```diff
@@ -0,0 +1,383 @@
...
+        # Use NVRX_LOG_DEBUG environment variable to determine log level
+        debug_enabled = os.environ.get("NVRX_LOG_DEBUG", "").lower() in ("1", "true", "yes", "on")
+        return logging.DEBUG if debug_enabled else logging.INFO
+
+    @classmethod
+    def get_process_name(cls, proc_name: Optional[str] = None) -> str:
+        return proc_name if proc_name is not None else str(os.getpid())
+
+    @classmethod
+    def get_log_to_stdop_cfg(cls) -> bool:
```
rhewett-nv: s/stdop/stdout

src/nvidia_resiliency_ext/shared_utils/log_manager.py
```diff
@@ -0,0 +1,383 @@
...
+        # In main script (launcher.py) or training subprocess
+        from nvidia_resiliency_ext.shared_utils.log_manager import setup_logger
+        logger = setup_logger()
+
+        # In subprocesses that need fresh logger setup
+        logger = setup_logger(force_reset=True)
+
+        # In other modules
+        import logging
+        logger = logging.getLogger(LogConfig.name)
```
rhewett-nv, line 335: Should import `LogConfig`

examples/inprocess/basic_example.py
```diff
@@ -246,8 +245,5 @@ def main(call_wrapper: Optional[inprocess.CallWrapper] = None):
     # example the Wrapper is applied to ``main()``, therefore logging needs to
     # be initialized and configured before the Wrapper is launched.
     args = parse_args()
-    logging.basicConfig(
```
rhewett-nv: Since the example is not using the new logger, we should keep the configuration the same.

examples/inprocess/optimal_example.py
```diff
@@ -216,10 +216,7 @@ def train(
 
 def main():
     args = parse_args()
-    logging.basicConfig(
```
rhewett-nv: Since the example is not using the new logger, we should keep the configuration the same.

src/nvidia_resiliency_ext/fault_tolerance/rank_monitor_server.py
```diff
@@ -96,6 +97,21 @@ def _create_restarter_sublogger(self):
...
+    """
+    Log a restart event that should always be visible, but only if restarter logging is enabled.
+    Args:
+        is_restarter_logger: Whether restarter logging is enabled
+        message: The message to log
+        *args, **kwargs: Additional arguments for logging
+    """
+    if is_restarter_logger:
+        # Get the nvrx logger directly
+        logger = logging.getLogger(LogConfig.name)
```
rhewett-nv: If it is a restarter logger, it should use the application logger, not the NVRx logger.  (Also need to verify the nested restarter logger for in-process works correctly (log with application, not with nvrx)

src/nvidia_resiliency_ext/shared_utils/log_manager.py
```diff
@@ -0,0 +1,383 @@
...
+import socket
+import sys
+from nvidia_resiliency_ext.shared_utils.log_distributed import (
+    DistributedLogHandler,
+    DynamicLogFormatter,
+)
+from typing import Optional
+
+
+class LogConfig:
```
rhewett-nv, line 104: We should be very explicit documenting that there is precisely _one_ `LogConfig` instance per process.

We may want to follow from Python logger at some point and have `setup_logging` take a string and return a singleton instance of `LogConfig` for that name.  We may want different instances for different components of NVRx in the same process.

src/nvidia_resiliency_ext/shared_utils/log_distributed.py
```diff
@@ -0,0 +1,399 @@
...
+        except (OSError, IOError, RuntimeError):
+            # Fallback to stderr if logging fails
+            sys.stderr.write(f"Log handler error: {record.getMessage()}\n")
+            sys.stderr.flush()
+
+    def _log_file_namer(self):
+        return f"rank_{self.rank_id}_{self.proc_name}.msg.{int(time.time()*1000)}"
+
+    def _cleanup_old_backup_files(self):
+        """Clean up old backup files, keeping only the most recent one's."""
```
rhewett-nv: Are these actually backups?  Or just old parts of the rotation?

src/nvidia_resiliency_ext/shared_utils/log_manager.py
```diff
@@ -0,0 +1,383 @@
...
+        allowing users to use all standard logging methods:
+        - logger.debug(message)
+        - logger.info(message)
+        - logger.warning(message)
+        - logger.error(message)
+        - logger.critical(message)
+        """
+        return self._logger
+
+    def __enter__(self):
```
rhewett-nv: Where is the context manager used?

src/nvidia_resiliency_ext/shared_utils/log_distributed.py
```diff
@@ -0,0 +1,399 @@
...
+    def _process_messages(self, output):
+        # Check for pending messages from other ranks
+        keep_processing = 50
+        msg_dict = {}
+        heap = []
+
+        while keep_processing:
+            if self._stop_event.is_set():
+                # Gives room for aggregator to catch up with writes
+                keep_processing -= 1
```
rhewett-nv, line 268: This will process for ~1.25s after the stop signal.  Is that enough time?  Flagging for a potential future change to make the time configurable.

src/nvidia_resiliency_ext/shared_utils/log_distributed.py
```diff
@@ -0,0 +1,399 @@
...
+
+        except FileNotFoundError as e:
+            # File was deleted between size check and read
+            sys.stderr.write(f"File not found during read {msg_file}: {e}\n")
+            sys.stderr.flush()
+            return
+        except (IOError, OSError) as e:
+            # File is being written by another process or other I/O error
+            # Log this as it might indicate a real problem
+            sys.stderr.write(f"IO error reading {msg_file}: {e}\n")
```
rhewett-nv, line 354: When should this bring the logger  or main process down?  I'm wary of printing indefinitely if there is a genuine I/O error that may not be recoverable.

src/nvidia_resiliency_ext/shared_utils/log_distributed.py
```diff
@@ -0,0 +1,399 @@
...
+        self._log_dict_queue[msg_file] = log_msg_q
+
+        # Update the position for this file
+        self._file_positions[msg_file] = file_size
+
+    def _cleanup_old_backup_files(self, msg_file: str):
+        """Clean up old backup files, keeping only the most recent one."""
+        # Find all backup files for this rank
+        parts_first = msg_file.split('.', 1)
+        parts_last = msg_file.rsplit('.', 1)
```
rhewett-nv, line 376: Edit: I think this actually is OK.  Can you document the expectations on the format of the `msg_file` name?  I initially assumed that it could only have one `.` but on subsequent review, I see otherwise.

src/nvidia_resiliency_ext/fault_tolerance/rank_monitor_server.py
```diff
@@ -510,9 +527,22 @@ def run(
...
+        # Set environment variables if provided
+        if env is not None:
+            for key, value in env.items():
+                os.environ[key] = value
+
+        # Set up the nvrx logger for subprocess
+        # Use force_reset=True to ensure fresh logger setup with correct rank info
+        from nvidia_resiliency_ext.shared_utils.log_manager import setup_logger
+
+        logger = setup_logger(force_reset=True, proc_name="rankmonsrv")
```
hexinw-nvidia: One feedback from @rhewett-nv is that "proc_name" is confusing against the LogConfig.name, which is the actual name of the logger module. Would it be more clear to change "proc_name" to "dist_log_file_prefix"?

src/nvidia_resiliency_ext/shared_utils/logger.py
namitdhameja: Will update comments to "When distributed logging is enabled, a separate aggregator service does log aggregation"

src/nvidia_resiliency_ext/shared_utils/logger.py
namitdhameja: While the dir is the same, the files are additive with a timestamp to prevent collisions.

src/nvidia_resiliency_ext/shared_utils/logger.py
namitdhameja: Now we do not delete the dir, the individual files are deleted once aggregation is performed on the file or it meets the max backup files.

.gitignore
namitdhameja: This is an old code, its already been moved out.

README.md
namitdhameja: The new changes does not include said file https://github.com/NVIDIA/nvidia-resiliency-ext/compare/main...hexinw-nvidia:nvidia-resiliency-ext:log

src/nvidia_resiliency_ext/inprocess/wrap.py
namitdhameja: Its included https://github.com/hexinw-nvidia/nvidia-resiliency-ext/blob/log/src/nvidia_resiliency_ext/inprocess/monitor_process.py#L145

examples/inprocess/basic_example.py
namitdhameja: That had been removed, this is an outdated file. Pls see https://github.com/hexinw-nvidia/nvidia-resiliency-ext/blob/log/examples/inprocess/basic_example.py

examples/inprocess/optimal_example.py
namitdhameja: Same comment. That had been removed, this is an outdated file.. 
https://github.com/hexinw-nvidia/nvidia-resiliency-ext/blob/log/examples/inprocess/optimal_example.py

src/nvidia_resiliency_ext/shared_utils/log_aggregator.py
namitdhameja: Its now "--en_chronological_ordering", will change to "-"

src/nvidia_resiliency_ext/fault_tolerance/rank_monitor_server.py
namitdhameja: okay

src/nvidia_resiliency_ext/shared_utils/log_manager.py
namitdhameja: okay

src/nvidia_resiliency_ext/shared_utils/log_manager.py
namitdhameja: Its fixed in latest https://github.com/hexinw-nvidia/nvidia-resiliency-ext/blob/log/src/nvidia_resiliency_ext/shared_utils/log_manager.py

src/nvidia_resiliency_ext/shared_utils/log_manager.py
namitdhameja: Will change the function name to drop kb.

src/nvidia_resiliency_ext/shared_utils/log_manager.py
namitdhameja: Its 4 backup plus 1 current, total 5. Will update comment

src/nvidia_resiliency_ext/shared_utils/log_manager.py
namitdhameja: Will change to stdout

src/nvidia_resiliency_ext/shared_utils/log_manager.py
namitdhameja: LogConfig is in the same file as LogManager; it includes some utility functions for parsing the env variables

examples/inprocess/basic_example.py
namitdhameja: Actually, internally, it calls setup_logger so it would override. I guess, if you want to use your own formatter then perhaps you should be ... have changed to address same.

src/nvidia_resiliency_ext/shared_utils/log_distributed.py
namitdhameja: will update comments

src/nvidia_resiliency_ext/shared_utils/log_distributed.py
namitdhameja: Its ~1.25 seconds after the service is stopped. The user has some measure of control by introducing sleep, I just added enough wait for UT to work fine. 
          $LAUNCHER_CMD $LAUNCHER_ARGS $WORKLOAD_CMD $WORKLOAD_ARGS
          sleep(<>)
          touch /nvrx_repo/stop 
          
I do agree with your comment that might have to revisit it.

src/nvidia_resiliency_ext/shared_utils/log_distributed.py
namitdhameja: rank_<id>_<prefix>.msg.<timestamp> [rank_0_test.msg.1755717568033]
The advantage of having timestamp is there is no naming conflict and after reaching the file limit, it just moves on to the new file. The aggregator, simply follows the file and cleans up after it. File rotation is not required in our use case.

src/nvidia_resiliency_ext/shared_utils/logger.py
namitdhameja: Logs are emitted in exception path

namitdhameja:
> Namit and I discussed over the design. There are three high level actions.
> 
> 1. We can move the aggregator function out of log_manager.py. The log_manager.py will be just NVRx logger functions.
[Namit] The aggregator is moved out to its own file

> 2. There will be multiple writers to a single temp per-rank log file (the training process and the monitor process). We need to look into how to do the log rotation correctly. But log rotation is not a P0 task. We can target this after basic logger change is in.
[Namit] Rotation is not a use case, being mindful of space utilization is. That is addressed by having file size limit, and deletion of processed files. 

> 3. We should make the design flexible that all ranks (on a same node) can write to the per-node local file directly. The log aggregator would become a simple rsync or tail service to stream the local file to LFS, just to reduce the pressure on Lustre performance in scale.
[Namit] You could run fluentbit as a tail service today and that would work just as well for aggregation use case, at the expense of ordering. That is independent of single or multiple files on the local fs.

src/nvidia_resiliency_ext/shared_utils/log_manager.py
namitdhameja: Nowhere, good catch. thx !

src/nvidia_resiliency_ext/shared_utils/logger.py
namitdhameja: Its not a rotate, its a new log file i.e. rank_<id>_<prefix>.msg.<timestamp>. The logs move to the new file with new timestamp. The aggregator checks for pattern, and keeps moving content to a central log file. It does give the advantages of a rotate though.

examples/inprocess/optimal_example.py
namitdhameja: okay

src/nvidia_resiliency_ext/shared_utils/log_manager.py
namitdhameja: Yes, will keep it in mind for future in case require singleton but vary per component. I don't think have an action item now.

namitdhameja (approved):

apaithankar (approved):
Big and welcome change! LGTM!

apaithankar (approved):

PR #128, opened 2025-07-22
[Build] Improvements to CUDA path detection during source installation

pramodk:
This is a minor thing but thought helpful to improve user experience.

### Summary

- If CUDA is installed and available in `$PATH`, it would be nice if build system could find it rather than throwing an error.
- Currently we just check for `/usr/local/cuda` and `CUDA_PATH` and then throw an error.
- In this PR, we try to check if `nvcc` present in the $PATH and try to determine the path of CUDA as it's typically done in build system like CMake: https://github.com/Kitware/CMake/blob/master/Modules/FindCUDA.cmake#L862

### As an example,

On DLCluster with CUDA setup in Conda environment:

```console
$ which nvcc
/home/scratch.prkumbhar_wwfo/software/x86_64/nvrx-july-2025/conda_envs/nvrx202507/bin/nvcc

$ pip install .
...
      File "/home/prkumbhar/workspace/repos/nvidia/nvidia-resiliency-ext/cupti_build.py", line 59, in build
          raise FileNotFoundError("cuda installation not found in /usr/local/cuda or $CUDA_PATH")
      FileNotFoundError: cuda installation not found in /usr/local/cuda or $CUDA_PATH
```

With this PR, it finds correct CUDA dir:

```console
  ...
  A setup.py file already exists. Using it.
  CUDA root found: /home/scratch.prkumbhar_wwfo/software/x86_64/nvrx-july-2025/conda_envs/nvrx202507/bin/../targets/x86_64-linux
```

### DIFF
--- a/cupti_build.py
+++ b/cupti_build.py
@@ -83,9 +83,7 @@ def get_cuda_path():
                 return match.group(1)
             else:
                 # fallback: get directory where nvcc is located
-                cuda_dir = os.path.dirname(os.path.dirname(nvcc_path))
-                if os.path.isdir(cuda_dir):
-                    return cuda_dir
+                return os.path.dirname(os.path.dirname(nvcc_path))
         except Exception:
             pass
 
--- a/cupti_build.py
+++ b/cupti_build.py
@@ -15,7 +15,9 @@
 
 import glob
 import os
-
+import re
+import shutil
+import subprocess
 from pybind11.setup_helpers import Pybind11Extension, build_ext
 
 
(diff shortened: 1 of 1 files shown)


cupti_build.py
```diff
@@ -43,6 +45,56 @@ def _skip_ext_build():
...
+            )
+            # match "#$ TOP=..." in output
+            match = re.search(r'#\$ TOP=([^\r\n]*)', result.stdout)
+            if match and os.path.isdir(match.group(1)):
+                return match.group(1)
+            else:
+                # fallback: get directory where nvcc is located
+                cuda_dir = os.path.dirname(os.path.dirname(nvcc_path))
+                if os.path.isdir(cuda_dir):
+                    return cuda_dir
```
hexinw-nvidia: Duplicate os.path.dirname? os.path.isdir will always be true. Right?
Is it fair to use NVCC directory as the CUDA directory?

hexinw-nvidia (approved):

cupti_build.py
pramodk: @hexinw-nvidia: you are right! `isdir` would be true even with broken links, very top level nvcc binary, etc. Updated PR.

> Is it fair to use NVCC directory as the CUDA directory?

This is just a last fallback, right? We can add more logic but I thought it's good to safeguard common pattern where we see failures (e.g. conda based installation, environment modules where cuda is loaded, etc).

hexinw-nvidia: Understood. Thanks for addressing the feedback.

PR #129, opened 2025-07-23
Add ability to temporarily disable hang protection

rhewett-nv:
This PR adds capability to temporarily disable hang protection on inprocess at targeted locations in workloads.

This is useful if there are known regions which do not release the GIL, which would normally trigger a restart due to soft or hard timeout.

This feature should be used only when the workload cannot be trivially corrected as it introduces some risk.  Non-hang related crashes are still able to be handled by inprocess restart.


hexinw-nvidia (approved):

src/nvidia_resiliency_ext/inprocess/monitor_process.py
```diff
@@ -252,10 +262,26 @@ def run(
                         hard_timeout = barrier_timeout
                         continue
 
-                    interruption = Interruption.HARD_TIMEOUT
-                    store.record_interrupted([InterruptionRecord(rank, interruption)])
-                    store.record_terminated_ranks([rank])
-                    terminate_process(target_process, termination_grace_time)
+                if timeouts_enabled:
```
rhewett-nv, line 262: Looks like a merge got messed up.  There are lines that should be removed.

hexinw-nvidia (approved):

src/nvidia_resiliency_ext/inprocess/wrap.py
```diff
@@ -390,6 +390,43 @@ def atomic(self):
         finally:
             self.atomic_lock.release()
 
+    @contextlib.contextmanager
+    def disable_hang_protection(self):
```
sbak5, line 450: There should be a set of interfaces to enable / disable timeouts for cases when user wants to disable them across multiple functions or a large segment of the codes which are not easy to be under a single `with`.

rhewett-nv: @sbak5 PTAL

sbak5: LGTM

docs/source/inprocess/usage_guide.rst
```diff
@@ -377,6 +377,55 @@ There are two methods to record progress:
...
+When the context manager exits (either normally or due to an exception), hang
+protection is automatically re-enabled.
+
+.. code-block:: python
+
+    def my_training_function(call_wrapper: CallWrapper):
+        # Normal operations are subject to hang protection
+        train_step()
+
+        # Disable hang protection for long-running data loading
```
apaithankar, line 462: I think this is a good first step, but I'm worried about cases seen recently where the storage cluster was being thrashed with `getattr()` calls and hung for a very long time.

Perhaps in the future we could add a way to temporarily extend the timeout for a section instead of disabling it completely.

apaithankar, line 409: I think this is a good first step, but I'm worried about cases seen recently where the storage cluster was being thrashed with `getattr()` calls and hung for a very long time.

Perhaps in the future we could add a way to temporarily extend the timeout for a section instead of disabling it completely.

apaithankar (approved):
LGTM

PR #133, opened 2025-07-29
Add logging to in-process rank assignment

rhewett-nv:
Users have requested better observability on rank state changes (e.g., inactive to active for hot spares) in In-process.


hexinw-nvidia (approved):

apaithankar (approved):
LGTM

PR #138, opened 2025-08-02
checkpointing: fix error propagation and add test

diegs:
Fixes error propagation during checkpoint saving
by using torch's DistWrapper to send the exception to the coordinator instead of killing the process
early. Also fixes call path to ensure necessary
operations happen in finally blocks.

Adds a test and slightly refactors the multiprocessing invocation to allow overriding the open() function with one that raises an exception. This enables the test. In the future we will factor our the filesystem operations into a delegate so it can be more easily mocked.

### DIFF
--- a/src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
+++ b/src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
@@ -37,6 +37,7 @@ import psutil
 import torch
 from torch import multiprocessing as mp
 from torch.distributed.checkpoint import FileSystemWriter
+from torch.distributed.checkpoint.api import WRAPPED_EXCEPTION, _wrap_exception
 from torch.distributed.checkpoint.filesystem import DEFAULT_SUFFIX, _StoragePrefix, _write_item
 from torch.distributed.checkpoint.metadata import Metadata
 
@@ -91,6 +92,7 @@ class FileSystemWriterAsync(FileSystemWriter):
     ):
         self.checkpoint_dir = path
         self.use_msc = kwargs.pop("use_msc", False)
+        self.open_file = kwargs.pop("open_file", open)  # for overriding in tests
 
         super().__init__(path, *args, **kwargs)
         if not self.single_file_per_rank:
@@ -204,10 +206,16 @@ class FileSystemWriterAsync(FileSystemWriter):
             3) arguments to that function in 1).
         """
         if not self.write_buckets:
-            return None, None, ()
+            return None, None, []
+        # TODO: refactor this for true storage fungibility.
+        if self.use_msc:
+            import multistorageclient as msc
+            open_file = msc.open
+        else:
+            open_file = self.open_file
         transform_list = [self.transforms] if hasattr(self, 'transforms') else []
         return (
-            partial(self.write_preloaded_data_multiproc, transform_list, self.use_msc),
+            partial(self.write_preloaded_data_multiproc, transform_list, self.use_msc, open_file),
             partial(self.preload_tensors, self.write_buckets, True),
             [torch.distributed.get_rank(), self.write_buckets, self.results_queue],
         )
@@ -238,6 +246,7 @@ class FileSystemWriterAsync(FileSystemWriter):
     def write_preloaded_data_multiproc(
         transform_list: List[_StorageWriterTransforms],
         use_msc: bool,
+        open_file: Callable,
         rank: int,
         write_buckets: List[WriteBucket],
         global_results_queue: mp.Queue,
@@ -287,7 +296,7 @@ class FileSystemWriterAsync(FileSystemWriter):
 
                 p_list.append(
                     ctx.Process(
-                        target=partial(FileSystemWriterAsync.write_preloaded_data, transform_list),
+                        target=partial(FileSystemWriterAsync.write_preloaded_data, transform_list, open_file),
                         kwargs=kwargs,
                     )
                 )
@@ -339,6 +348,7 @@ class FileSystemWriterAsync(FileSystemWriter):
     @_disable_gc()
     def write_preloaded_data(
         transform_list: List[_StorageWriterTransforms],
+        open_file: Callable,
         local_proc_idx: int,
         write_bucket: WriteBucket,
         results_queue: mp.SimpleQueue,
@@ -378,12 +388,6 @@ class FileSystemWriterAsync(FileSystemWriter):
                 assert len(transform_list) <= 1
                 write_fn = partial(_write_item, *transform_list)
 
-            if use_msc:
-                import multistorageclient as msc
-
-                open_file = msc.open
-            else:
-                open_file = open
             with open_file(file_name, "wb") as stream:
                 for write_item, data in bytes_data:
                     local_results.append(
@@ -421,7 +425,7 @@ class FileSystemWriterAsync(FileSystemWriter):
         """Write all items from ``plan``."""
         raise NotImplementedError('write_data not implemented for FileSystemWriterAsync')
 
-    def retrieve_write_results(self) -> List[WriteResult]:
+    def retrieve_write_results(self) -> Union[List[WriteResult], WRAPPED_EXCEPTION]:
         """
         Turn the latest dict including write results from `self.results_queue`
             into a single results lists. Includes error check.
@@ -438,16 +442,19 @@ class FileSystemWriterAsync(FileSystemWriter):
             try:
                 write_results_or_exc = self.results_queue.get_nowait()
             except queue.Empty:
-                raise RuntimeError('results_queue should not be empty')
+                return _wrap_exception(RuntimeError('results_queue should not be empty'))
 
         if isinstance(write_results_or_exc, Exception):
-            raise RuntimeError(f'Worker failure: {write_results_or_exc}') from write_results_or_exc
+            try:
+                raise RuntimeError(f'Worker failure: {write_results_or_exc}') from write_results_or_exc
+            except Exception as e:
+                return _wrap_exception(e)
         write_results: dict = write_results_or_exc
         if len(write_results) != len(self.write_buckets):
-            raise RuntimeError(
+            return _wrap_exception(RuntimeError(
                 f'Incomplete worker results (expected {len(self.write_buckets)},'
                 f' got {len(write_results)}. This probably indicates a worker failure.'
-            )
+            ))
         return list(chain.from_iterable(write_results.values()))
 
     def prepare_decentralized_global_plan(self, local_plan: SavePlan) -> SavePlan:
@@ -484,7 +491,7 @@ class FileSystemWriterAsync(FileSystemWriter):
 
             # storage_meta was introduced since PyTorch 2.4
             if "storage_meta" in inspect.signature(Metadata).parameters:
-                metadata.stoage_meta = self.storage_meta()
+                metadata.storage_meta = self.storage_meta()
 
             path = os.path.join(self.checkpoint_dir, ".metadata")
 
--- a/tests/checkpointing/unit/conftest.py
+++ b/tests/checkpointing/unit/conftest.py
@@ -20,6 +20,8 @@ import pytest
 from . import TempNamedDir
 from .test_utilities import Utils
 
+from nvidia_resiliency_ext.checkpointing.async_ckpt.core import AsyncCallsQueue
+
 
 @pytest.fixture(scope="session")
 def tmp_path_dist_ckpt(tmp_path_factory) -> Path:
@@ -37,3 +39,9 @@ def tmp_path_dist_ckpt(tmp_path_factory) -> Path:
 
     else:
         yield tmp_dir
+
+@pytest.fixture
+def async_queue():
+    async_queue = AsyncCallsQueue()
+    yield async_queue
+    async_queue.close()
--- a/tests/checkpointing/unit/test_async_save.py
+++ b/tests/checkpointing/unit/test_async_save.py
@@ -13,23 +13,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import torch
+import pytest
 
 from nvidia_resiliency_ext.checkpointing.async_ckpt.torch_ckpt import TorchAsyncCheckpoint
 
 from . import TempNamedDir
-from .test_utilities import TestModel, Utils
+from .test_utilities import Model, Utils
 
-
+@pytest.mark.skip(reason='broken test, needs reviving')
 class TestAsyncSave:
-    def setup_method(self, method):
-        Utils.set_world_size(1)
-
-    def teardown_method(self, method):
-        Utils.set_world_size()
 
     def test_async_is_equivalent_to_sync(self, tmp_path_dist_ckpt):
         Utils.initialize_distributed()
-        model = TestModel((1024, 1024), 10)
+        model = Model((1024, 1024), 10)
         ckpt_impl = TorchAsyncCheckpoint()
         state_dict = model.state_dict()
         with (
--- a/tests/checkpointing/unit/test_async_writer.py
+++ b/tests/checkpointing/unit/test_async_writer.py
@@ -17,8 +17,13 @@ import pickle
 from copy import deepcopy
 from dataclasses import fields
 
+import os
+from typing import Any, IO
+
+import pytest
 import torch
 from torch.distributed.checkpoint import (
+    CheckpointException,
     DefaultLoadPlanner,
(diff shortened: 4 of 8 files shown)


diegs:
With this MR, running pytest for the entire `tests/checkpointing/unit` directory passes. In subsequent PRs we will improve and expand these tests.

tests/checkpointing/unit/test_async_save.py
```diff
@@ -13,23 +13,19 @@
...
+import pytest
 
 from nvidia_resiliency_ext.checkpointing.async_ckpt.torch_ckpt import TorchAsyncCheckpoint
 
 from . import TempNamedDir
-from .test_utilities import TestModel, Utils
-
+from .test_utilities import Model, Utils
 
+@pytest.mark.skip(reason='broken test, needs reviving')
```
aartibasant: why are we skipping tests?

tests/checkpointing/unit/test_async_writer.py
```diff
@@ -46,16 +60,18 @@ def get_async_save_request(self, writer, save_state_dict_ret) -> AsyncRequest:
 
         def finalize_fn():
             """Finalizes async checkpointing and synchronizes processes."""
-            save_state_dict_async_finalize(*save_state_dict_ret)
-            torch.distributed.barrier()
+            try:
```
aartibasant: do the users have to have the try catch? I thought you already added that in our finalize functionality. Why is the test re-defining this?

tests/checkpointing/unit/test_async_writer.py
```diff
@@ -36,8 +41,17 @@
...
+from logging import getLogger
+import logging
+
+def mock_open(
+        self,
+        path: str,
+        mode: str = "rb",
+    ) -> IO[Any]:
+    """Raises an error on worker #2 during storage save"""
+    raise OSError('worker #2 critical failure')
```
aartibasant: This "open" will raise exception on all workers. Not just worker # 2 right? You restrict worker # 2 in the caller. Not in this function itself. So the error comment is inaccurate.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -204,10 +206,16 @@ def get_save_function_and_args(self) -> Tuple[Optional[Callable], Optional[Calla
             3) arguments to that function in 1).
         """
         if not self.write_buckets:
-            return None, None, ()
+            return None, None, []
+        # TODO: refactor this for true storage fungibility.
```
aartibasant: We can remove this TODO. We will cleanly address it when we get to storage fungibility

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -421,7 +425,7 @@ def write_data(self, plan: SavePlan, planner: SavePlanner) -> Future[List[WriteR
         """Write all items from ``plan``."""
         raise NotImplementedError('write_data not implemented for FileSystemWriterAsync')
 
-    def retrieve_write_results(self) -> List[WriteResult]:
+    def retrieve_write_results(self) -> Union[List[WriteResult], WRAPPED_EXCEPTION]:
```
aartibasant, line 430: Update function comments on update return types and condition in which exception is returned

tests/checkpointing/unit/test_async_save.py
```diff
@@ -13,23 +13,19 @@
...
 class TestAsyncSave:
-    def setup_method(self, method):
-        Utils.set_world_size(1)
-
-    def teardown_method(self, method):
-        Utils.set_world_size()
 
     def test_async_is_equivalent_to_sync(self, tmp_path_dist_ckpt):
         Utils.initialize_distributed()
-        model = TestModel((1024, 1024), 10)
```
aartibasant, line 33: Why is this PR updating TestModel to Model? I do not understand in this context

tests/checkpointing/unit/test_async_writer.py
```diff
@@ -125,41 +140,59 @@ def test_async_is_equivalent_to_sync(self, tmp_path_dist_ckpt):
                     loaded_sync_state_dict[key], state_dict[key]
                 ), f"Mismatch for key '{key}' between async checkpoint and original state_dict."
 
-    def test_cached_metadata(self, tmp_path_dist_ckpt):
+    def test_errors_are_reported(self, tmp_path_dist_ckpt, async_queue):
```
aartibasant, line 148: Is async_queue setup and teardown needed between tests?
The point is that even our training loop may get an exception from CP error and may choose to ignore and continue. In that case async_queue need not be setup again.

tests/checkpointing/unit/test_async_save.py
diegs: This test is currently broken, but I want to change CI (separate repo) to run all the tests in the directory. Currently it only runs 2 files.

tests/checkpointing/unit/test_async_save.py
diegs: Sorry, should have called this out. Any class name that starts with `Test` causes pytest to try and run test cases in it, and throws a warning. Renaming it to `Model` simply to remove that noise. Could also call it `ToyModel` or similar.

tests/checkpointing/unit/test_async_writer.py
diegs: The `finalize_fn()` is user-provided. In this case I think the barrier was placed here for testing purposes, so as is true with any barrier we want to make sure every rank always goes through it.

Another way to accomplish this is to put the barrier in the test case itself. But the original tests did not do that.

tests/checkpointing/unit/test_async_writer.py
diegs: I can change it to a session-scoped fixture. However that reduces test isolation. If we want to test the above case (ignore exception, keep using the async loop) I think we should make that an explicit test case. I can add it to the list if you agree.

tests/checkpointing/unit/test_async_writer.py
aartibasant: To make sure we do not rely on users providing this `try-catch` may I propose the following:

1. User defines the `finalize_fn():`
2. However, the finalize function is executed by us in https://github.com/NVIDIA/nvidia-resiliency-ext/blob/main/src/nvidia_resiliency_ext/checkpointing/async_ckpt/core.py#L103
3. You have already triggered the `try-catch` in the caller
So I think this should work with you modifying the user defined finalize_fn.

The reason I think this is important is we cannot rely on users writing their finalze functions with the expected try-catch. So I want to make sure in our unit tests, we correctly mock that behavior and have users write simpler callables. When we invoke these functions, this PR is doing the necessary error propogation.

Suggestion is to remove the `try-catch` from this `finalize_fn `

tests/checkpointing/unit/test_async_writer.py
```diff
@@ -46,16 +58,27 @@ def get_async_save_request(self, writer, save_state_dict_ret) -> AsyncRequest:
...
+        state_dict,
+        planner,
+        async_queue: AsyncCallsQueue,
+        thread_count=1,
+        caching=False,
+        open_file=open,
     ):
         """Performs an asynchronous model checkpoint save."""
-        writer = FileSystemWriterAsync(checkpoint_dir, thread_count=thread_count)
+        writer = FileSystemWriterAsync(
```
aartibasant, line 76: Can you write a test to make sure there is no backward compatibility issue in case we do not define "open_file"?
Our users may not have this API extension.
I think your code looks fine. Just want to make sure we have a backward compatible test.

tests/checkpointing/unit/test_async_writer.py
aartibasant: Yes agreed. If this setup tear down is improving test stability and simplicity to deug, I am okay with this change.
But in that case, I request an explicit test where an exception does occur and we need not setup teardown the persistent worker or async_queues and then training simply continues to the next CP write.

tests/checkpointing/unit/test_async_save.py
```diff
@@ -12,24 +12,21 @@
...
 
 from . import TempNamedDir
-from .test_utilities import TestModel, Utils
+from .test_utilities import Model, Utils
 
 
+@pytest.mark.skip(reason='broken test, needs reviving')
 class TestAsyncSave:
-    def setup_method(self, method):
-        Utils.set_world_size(1)
```
aartibasant, line 26: Shouldn't `Utils.initialize_distributed()` be in setup?
Could this be why this unit test is failing?

tests/checkpointing/unit/test_async_writer.py
diegs: No longer applies with updated approach

tests/checkpointing/unit/test_async_writer.py
diegs: The existing test cases, for the happy path and happy path with MSC, do not pass `open_file`, which I believe is sufficient coverage for that the new argument is indeed optional.

tests/checkpointing/unit/test_async_writer.py
diegs: I made the async queue session scoped. This mean the same queue is persisted across all tests cases that use it, ensuring that it gracefully handles re-use.

tests/checkpointing/unit/test_async_save.py
diegs: `Utils.set_world_size()` implicitly calls `initialize_distributed()`. There is something else going on with this test case. It is also redundant to the `test_async_writer.py` test cases.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -438,15 +444,22 @@ def retrieve_write_results(self) -> List[WriteResult]:
             try:
                 write_results_or_exc = self.results_queue.get_nowait()
             except queue.Empty:
-                raise RuntimeError('results_queue should not be empty')
+                return _wrap_exception(RuntimeError('results_queue should not be empty'))
 
         if isinstance(write_results_or_exc, Exception):
-            raise RuntimeError(f'Worker failure: {write_results_or_exc}') from write_results_or_exc
```
aartibasant, line 444: Why don't you just 

return _wrap_exception(RuntimeError(...))

tests/checkpointing/unit/test_async_writer.py
```diff
@@ -81,11 +101,10 @@ def load_checkpoint(self, checkpoint_dir, state_dict):
         )
         return state_dict
 
-    def test_async_is_equivalent_to_sync(self, tmp_path_dist_ckpt):
+    def test_async_is_equivalent_to_sync(self, tmp_path_dist_ckpt, async_queue):
```
aartibasant, line 104: Can you please expalin to me again how this async queue object is created and why do you need to pass it instead of creating it in the test?

I understand that the async_queue is not "closed" in any of the tests.
If at the end of these tests, you did async_queue.close() do you still need to make this change of passing an async_queue?
I suspect the hangs you were seeing was probably because async_queue was not beeing closed in the tests.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
diegs: I believe I wrote the answer to this before; you can't use `from` to do exception chaining without a `raise`.

tests/checkpointing/unit/test_async_writer.py
diegs: The `async_queue` is a session scoped pytest fixture. It is created once the first time it is used, reused across all tests that need it, and then closed after the tests conclude.

I could create it in the test and close it at the end, but this is more idiomatic for pytest (and allows reusing the same queue across all the tests to make sure it is able to reused, as we discussed before).

The hangs were indeed due to not closing the queue. If the queue is not explicitly closed then it closes itself in its destructor. However if not explicitly closed in these tests, pytest hangs indefinitely after the test cases finish. My understanding is that pytest attempts to collect test results from all subprocesses, so it probably has a hook to keep track of those subprocesses, and by keeping a reference to the subprocess it ironically prevents the queue from being destructed. Explicitly closing the queue kills the persistent subprocess and allows pytest to shutdown cleanly.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
aartibasant: Got it

tests/checkpointing/unit/test_async_save.py
aartibasant: But arn't we fixing this test via this PR?

tests/checkpointing/unit/conftest.py
```diff
@@ -37,3 +39,10 @@ def tmp_path_dist_ckpt(tmp_path_factory) -> Path:
 
     else:
         yield tmp_dir
+
+
+@pytest.fixture(scope="session")
```
aartibasant: I plan to add a unit test that restarts this Async_queue due to inprocess restarts.
In that case this fixture may have to extend with somethng like bwlow:
Writing this as a pointer to myself when we extend this unit test. Not asking you to do this in this PR.


```
class AsyncCallsQueueTestWrapper:
    def __init__(self):
       if self._queue  is None:
               self._queue = AsyncCallsQueue(persistent=True)
   
    
    def get(self):
        return self._queue

    def restart(self):
        # Clean up old queue (sync or async as needed)
        self._queue.close()
        self._queue = AsyncCallsQueue(persistent=True)

    def close(self):
        self._queue.close()

@pytest.fixture(scope="session")
def async_queue():
    async_queue = AsyncCallsQueueTestWrapper()
    yield async_queue
    async_queue.close()
```

Then the code that exercises the in-process restart will have to explicitly call the `async_queue.restart()` method.

aartibasant: @diegs I do not see this being done in mCore suite. Why do you suggest we need a unque async queue for all tests in NVRx. I suspect this will complicate restart unit tests which we plan to expand on.

cc: @sbak5 for review on this too.

diegs: @aartibasant I think this discussion has gotten confusing, we can talk face to face if that is easier.

Previously, each test case would directly create an `AsyncCallsQueue()` and never close it.

When you changed the default to `persistent=True`, this causes a hang in `pytest` because it is waiting for the persistent subprocess to end, but that subprocess does not end (destruction does not happen).

Adding a fixture is a more idiomatic way to have a resource that can be used for multiple tests. Initially it was a fresh queue for each test that calls for it. After our previous discussion, I made it a single queue for the entire test session. In the happy case, we expect that the queue is reusable.

However, you are free to write tests that create their own `AsyncCallsQueue` directly. You never *have* to use the fixture. For specific tests that want to do testing of the queue itself, you would do that. For tests where the queue is simply a dependency, it's cleaner to extract it to a fixture.

tests/checkpointing/unit/test_async_writer.py
aartibasant: @diegs : I patched this PR and can confirm the following:
1. I deleted the session scoped async_queue in the conftest.py
2. In every test, I created a async_queue = AsyncCallsQueue()
3. At the end of the test, I made sure that async_queue.close() is called

This resolved the hang that you were seeing
Basically these tests are not closing the async_queue which is triggering the hanging child-processes.

Instead of doing this in a scoped, session, we should rather fix the unit tests and cleanly close the queue in tests.

The reason I am suggesting this is that there will be a future PR inw hich we will be testing functionality of CP aborts. In those case, the tests will re-start asyncCallsQueue. Having thgis scoped session in those cases will be very confusing.

diegs: @aartibasant yes, we are going around in circles here a bit. I identified this fix before (you must close the async queue in every test). The fixture (session scoped or not) just abstracts this into the pytest framework, so the test code can focus on the system under test and not the test dependencies, per testing best practices. Specifically, tests can just include the `async_queue` as an argument, let the fixture handle the setup and closing, and not worry about it.

I'm also fine if we want to revert to explicitly creating it and closing it manually in every test. I think a fixture is more ergonomic though. The effect is the same either way.

tests/checkpointing/unit/conftest.py
diegs: I changed it to a non-session-scoped fixture per our offline discussion.

aartibasant (approved):

PR #145, opened 2025-08-20
Add enhanced GPU and NVL health checks with automatic chaining

hexinw-nvidia:
- Refactor GPUHealthCheck to support device-specific health monitoring
- Add new NVLHealthCheck class for NVLink health validation
- Implement automatic health check chaining in Wrapper class
- Create ChainedGPUHealthCheck and ChainedNVLHealthCheck for in-process use
- Update trace collector to use new GPU health check API
- Add related test coverage for new health check functionality

### DIFF
--- a/src/nvidia_resiliency_ext/inprocess/wrap.py
+++ b/src/nvidia_resiliency_ext/inprocess/wrap.py
@@ -19,6 +19,7 @@ import datetime
 import functools
 import gc
 import logging
+import os
 import pathlib
 import sys
 import threading
@@ -36,7 +37,7 @@ from .completion import Completion
 from .compose import Compose
 from .exception import HealthCheckError, InternalError
 from .finalize import Finalize
-from .health_check import HealthCheck
+from .health_check import ChainedGPUHealthCheck, ChainedNVLHealthCheck, HealthCheck
 from .initialize import Initialize
 from .monitor_process import MonitorProcess
 from .monitor_thread import MonitorThread, RankShouldRestart, reraise_if_unraisable
@@ -95,7 +96,10 @@ class Wrapper:
         initialize: Rank-local initialize.
         abort: Asynchronously aborts execution.
         finalize: Rank-local finalize.
-        health_check: Rank-local health check.
+        health_check: Rank-local health check executed before restart. This health check is
+            automatically enhanced with GPU and NVL health validation to ensure comprehensive
+            system health before restart.
+
         rank_assignment: Reassigns ranks, computes the new world size and
             specifies which ranks are calling the wrapped function.
         rank_filter: (DEPRECATED) Specifies ranks actively calling the wrapped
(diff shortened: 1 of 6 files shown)


src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -621,3 +594,135 @@ def _perform_health_check(self) -> bool:
...
+                        self.log.warning(
+                            f"GPU {device_index}: NVL link {link_id} is in DISABLED state"
+                        )
+                        return False
+
+                except self.pynvml.NVMLError as e:
+                    # Link might not exist or be accessible, which is normal
+                    # Only log if it's not a "not supported" error
+                    if "not supported" not in str(e).lower():
+                        self.log.debug(
```
rhewett-nv: Is debug the right level here?  The GPU health checks are warning.

hexinw-nvidia: Thanks for catching it. This should be a warning.

sbak5:
The health checks simply return True or False but the inprocess loop takes action only when the helathcheck raises exceptions so even if it says False, it doesn't do any action for the failed health check

src/nvidia_resiliency_ext/inprocess/wrap.py
```diff
@@ -523,8 +561,8 @@ def __call__(self, fn, args, kwargs):
                             raise finalize_ex from term_ex
 
                         try:
-                            if wrapper.health_check is not None:
-                                wrapper.health_check(state.freeze())
+                            if wrapper.restart_health_check is not None:
+                                wrapper.restart_health_check(state.freeze())
```
sbak5, line 565: This will not raise any exception because the underlying routine doesn't do that.

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -330,26 +308,28 @@ def _perform_health_check_single_gpu(self, gpu_id: int) -> bool:
...
-                    f"GPU {gpu_id}: Operating at reduced capacity. Drain existing work and reset the GPU."
+                    f"GPU {device_id}: Operating at reduced capacity. Drain existing work and reset the GPU."
                 )
                 return False
             else:
-                self.log.warning(f"GPU {gpu_id}: Unknown recovery action status: {recovery_action}")
+                self.log.warning(
+                    f"GPU {device_id}: Unknown recovery action status: {recovery_action}"
+                )
                 return False
```
sbak5, line 334: Raising exception may be required.

hexinw-nvidia: The healthcheck API returns True/False. The downstream module then decides what to do with the information.
In InProcess, it raises HealthCheckError exception when the GPU healthcheck fails.

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -621,3 +594,135 @@ def _perform_health_check(self) -> bool:
...
+                [self.device_index]
+                if self.device_index is not None
+                else range(self.pynvml.nvmlDeviceGetCount())
+            )
+
+            # Check all specified devices
+            all_healthy = True
+            for device_id in devices_to_check:
+                if not self._check_nvl_links_for_device(device_id):
+                    all_healthy = False
```
sbak5, line 678: This can be very aggressive to return False simply on NVL health checks. 
It's fine for this initial version but needs improvement to convey more detailed info. 
because a single link failure doesn't interrupt workload. And many of jobs are not affected at all.

apaithankar (commented):
LGTM

PR #151, opened 2025-08-26
checkpoint test utils: fix rank/world size

diegs:
`WORLD_SIZE` is set to the correct value via torchrun. Using the cuda device count is not correct.

Additionally, `LOCAL_RANK` is only equal to `RANK` on a single node. The `rank` variable is used as a global rank, so correctly set it using `RANK`.

This is required to run the tests on CI on different shaped machines, such as ARM.


sbak5:
the test is supposed to run in a single node. I think the previous code should work but not sure why it doesn't work in CI for ARM.

diegs:
@sbak5 in the logs @hexinw-nvidia shared, there were only 2 gpus on the node. So the other 6 processes crashed during distributed init trying to set their device to an invalid ID

hexinw-nvidia (approved):

diegs:
@hexinw-nvidia I am not sure why the downstream gitlab-ci test failed, and I can't seem to retrigger it. When you have a chance can you please help me understand what's wrong with it? All I can see is `trigger_tests` failed with `no permissions to trigger downstream pipeline`. Thanks!

hexinw-nvidia:
@diegs It failed because the downstream CI is not triggered. I have restarted the CI.

aartibasant (approved):

PR #154, opened 2025-08-28
Ability to abort an async CP process

aartibasant:
This PR extends the ability to abort the async CP workers.

Problem Statement: In case of fault tolerance situations like inprocess restarts, we want all ranks to have the ability to abort the async CP workers and restart them with inprocess. This is to ensure explicit closure of async CP workers during inprocess restarts.

The Solution: The AsyncCallsQueue that manages the async CP workers is made a singleton so that every rank can have only 1 such process. With CP abort, we make sure to shutdown the async process so that a subsequent call to save CP will restart a new process. This functionality is extended to the inprocess Abort routines.

Unit test: A unit test is added to that takes a clean CP, aborts the CP processes and the takes a 2nd CP. We validate CP state before and after the abort.

### DIFF
--- a/src/nvidia_resiliency_ext/inprocess/abort.py
+++ b/src/nvidia_resiliency_ext/inprocess/abort.py
@@ -188,7 +188,7 @@ class AbortTransformerEngine(Abort):
         return state
 
 
-class AbortCheckpoint(Abort):
+class AbortPersistentCheckpointProcesses(Abort):
     r'''
     Aborts Async Checkpoint processes
 
--- a/src/nvidia_resiliency_ext/inprocess/abort.py
+++ b/src/nvidia_resiliency_ext/inprocess/abort.py
@@ -22,6 +22,7 @@ import os
 import torch
 
 from nvidia_resiliency_ext.attribution.trace_analyzer.trace_collector import TorchFRTraceCollector
+from nvidia_resiliency_ext.checkpointing.async_ckpt.core import abort_nvrx_checkpoint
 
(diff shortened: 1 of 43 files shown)


src/nvidia_resiliency_ext/inprocess/abort.py
```diff
@@ -185,3 +186,14 @@ def __call__(self, state: FrozenState) -> FrozenState:
             te_fp8.FP8GlobalStateManager.reset()
 
         return state
+
+
+class AbortCheckpoint(Abort):
```
rhewett-nv: The PR looks great.  My only nit is that `AbortCheckpoint` kind of implies that it aborts an in-progress checkpoint.  Can we change this to something like `AbortPersistentCheckpointProcesses` or something?  I don't mind if it is wordy..

rhewett-nv (approved):

rhewett-nv (approved):

src/nvidia_resiliency_ext/inprocess/abort.py
```diff
@@ -188,7 +188,7 @@ def __call__(self, state: FrozenState) -> FrozenState:
         return state
 
 
-class AbortCheckpoint(Abort):
+class AbortPersistentCheckpointProcesses(Abort):
```
aartibasant, line 191: @rhewett-nv Thanks for the review. Addressed your comment.

hexinw-nvidia (approved):

PR #155, opened 2025-08-29
Use iteration count and NVRX_FR_TRACE_PATH env for fr trace path

sbak5:
Use iteration count for FR trace path not to override existing traces across restarts
Also, use `NVRX_FR_TRACE_PATH` when trace path is not passed to `AbortTorchDistributed`

### DIFF
--- a/src/nvidia_resiliency_ext/inprocess/abort.py
+++ b/src/nvidia_resiliency_ext/inprocess/abort.py
@@ -88,9 +88,7 @@ class AbortTorchDistributed(Abort):
         Args:
             torch_fr_trace_path: Path to collect PyTorch Flight Recorder traces.
         '''
-        if torch_fr_trace_path is None:
-            torch_fr_trace_path = os.environ.get('NVRX_FR_TRACE_PATH', None)
-        self.torch_fr_trace_path = torch_fr_trace_path
+        self.torch_fr_trace_path = torch_fr_trace_path or os.environ.get('NVRX_FR_TRACE_PATH', None)
 
     @staticmethod
     def shutdown_all_process_group_backends():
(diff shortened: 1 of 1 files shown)


rhewett-nv (dismissed):

src/nvidia_resiliency_ext/inprocess/abort.py
```diff
@@ -90,6 +88,8 @@ def __init__(self, torch_fr_trace_path: str = None):
         Args:
             torch_fr_trace_path: Path to collect PyTorch Flight Recorder traces.
         '''
+        if torch_fr_trace_path is None:
+            torch_fr_trace_path = os.environ.get('NVRX_FR_TRACE_PATH', None)
```
hexinw-nvidia: Nit: A more pythonic code would be:

torch_ft_trace_path = torch_ft_trace_path or os.environ.get("NVRX_FR_TRACE_PATH")

src/nvidia_resiliency_ext/inprocess/abort.py
```diff
@@ -136,27 +136,28 @@ def _check_fr_env():
                 except ValueError:
                     env_value_int = 0
```
hexinw-nvidia, line 138: Line 134-138 can be deleted.

src/nvidia_resiliency_ext/inprocess/abort.py
```diff
@@ -136,27 +136,28 @@ def _check_fr_env():
...
             return True
 
         if _check_fr_env() is True:
             if self.torch_fr_trace_path is None:
                 return
-            if not os.path.exists(self.torch_fr_trace_path):
-                os.makedirs(self.torch_fr_trace_path)
-            trace_analyzer = TorchFRTraceCollector(self.torch_fr_trace_path)
+            trace_path = os.path.join(self.torch_fr_trace_path, f'iter_{state.iteration}')
+            if not os.path.exists(trace_path):
```
hexinw-nvidia: os.path.exists is not necessary.

src/nvidia_resiliency_ext/inprocess/abort.py
sbak5: env var can be a value not convertible to int

hexinw-nvidia: We should flag the error to the user as it is a config error. Better than suppressing the error and quietly using 0 underneath.

hexinw-nvidia (dismissed):
LG. Thanks for making the change.

rhewett-nv (approved):

PR #156, opened 2025-08-29
Named change NVRX_DIST_LOG_DIR to NVRX_NODE_LOCAL_TMPDIR

hexinw-nvidia:
.

### DIFF
--- a/docs/source/shared_utils/api.rst
+++ b/docs/source/shared_utils/api.rst
@@ -135,8 +135,8 @@ Log Manager Properties
 
    * - Property
      - Description
-   * - ``distributed_logging_enabled``
-     - Boolean indicating whether distributed logging is enabled
+   * - ``node_local_tmp_logging_enabled``
+     - Boolean indicating whether node local temporary logging is enabled
    * - ``workload_rank``
      - Integer representing the workload rank
    * - ``workload_local_rank``
--- a/docs/source/shared_utils/api.rst
+++ b/docs/source/shared_utils/api.rst
@@ -3,6 +3,9 @@ API Reference
 
 This section provides detailed API documentation for the NVRx Shared Utilities, focusing on the logging system.
 
+.. note::
+   For configuration options, environment variables, examples, and usage guides, see :doc:`config_reference`.
+
 Log Manager
 ----------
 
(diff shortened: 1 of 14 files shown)


namitdhameja (approved):

docs/source/shared_utils/config_reference.rst
```diff
@@ -0,0 +1,238 @@
...
+   :header-rows: 1
+
+   * - Variable
+     - Type
+     - Default
+     - Description
+   * - ``NVRX_NODE_LOCAL_TMPDIR``
+     - string
+     - None
+     - Directory for temporary log files. When set, enables distributed logging mode.
```
rhewett-nv: Should emphasize that it should not be placed on a filesystem that cannot handle lots of writes from all nodes.

docs/source/shared_utils/api.rst
```diff
@@ -0,0 +1,167 @@
...
+     - Directory path for temporary log files. When set, enables distributed logging mode.
+   * - ``NVRX_LOG_DEBUG``
+     - string
+     - Set to "1", "true", "yes", or "on" to enable DEBUG level logging. Default: INFO level.
+   * - ``NVRX_LOG_TO_STDOUT``
+     - string
+     - Set to "1" to log to stdout instead of stderr. Default: stderr.
+   * - ``NVRX_LOG_MAX_FILE_SIZE_KB``
+     - integer
+     - Maximum size of temporary message files in KB before rotation. Default: 10.
```
apaithankar: IIUC, once a file reaches 10 KB it will be rotated? Usually the syslogs are rotated after 1GB of messages. 10KB seems very low for long pre-training runs.

docs/source/shared_utils/api.rst
```diff
@@ -0,0 +1,167 @@
...
+   * - ``NVRX_LOG_MAX_LOG_FILES``
+     - integer
+     - Maximum number of log files to keep per rank. Default: 4.
```
apaithankar: Why are we deleting _any_ logs? Let the user decide what to keep and delete and keep everything by default?

docs/source/shared_utils/api.rst
```diff
@@ -0,0 +1,167 @@
...
+   * - Variable Name
+     - Type
+     - Description
+   * - ``NVRX_NODE_LOCAL_TMPDIR``
+     - string
+     - Directory path for temporary log files. When set, enables distributed logging mode.
+   * - ``NVRX_LOG_DEBUG``
+     - string
+     - Set to "1", "true", "yes", or "on" to enable DEBUG level logging. Default: INFO level.
+   * - ``NVRX_LOG_TO_STDOUT``
```
apaithankar: does this work like `tee` or logs are redirected to file only?

docs/source/shared_utils/api.rst
```diff
@@ -0,0 +1,167 @@
...
+Log Manager Properties
+---------------------
+
+.. list-table:: LogManager Properties
+   :widths: 30 70
+   :header-rows: 1
+
+   * - Property
+     - Description
+   * - ``distributed_logging_enabled``
```
apaithankar: Can we explain what *distributed logging* here means? It's not clear.

docs/source/shared_utils/logger_guide.rst
```diff
@@ -0,0 +1,164 @@
+NVRx Logger Guide
+=================
+
+The NVRx Logger is a sophisticated distributed logging system designed specifically for multi-node, multi-rank training workloads. It provides intelligent log aggregation, rank-aware formatting, and automatic adaptation between distributed and regular logging modes.
```
apaithankar: Can we explain what exactly we mean by *distributed logging system*? How is it different from standard logging in PyTorch?

rhewett-nv (dismissed):
LGTM, thanks!

apaithankar (dismissed):
Some minor comments, but otherwise LGTM

docs/source/shared_utils/api.rst
hexinw-nvidia: Good catch. It is a document error. The default is 10MB, which is a reasonable file size.

docs/source/shared_utils/api.rst
hexinw-nvidia: These are the settings for user to decide how many temp files they want to keep before deleting the temp file.

docs/source/shared_utils/api.rst
hexinw-nvidia: I didn't get the question. NVRX_LOG_TO_STDOUT sets up the log handler to go to stdout instead of stderr.

apaithankar: The question applies to both cases (stdout going to stderr or stdout).. When the log are written to stdout/err, will they also still be written to the log files or only the stdout/err?

docs/source/shared_utils/api.rst
hexinw-nvidia: Agreed. Fixed the doc. Please check.

docs/source/shared_utils/logger_guide.rst
hexinw-nvidia: Ack. Please check again.

anjalibshah (approved):

docs/source/shared_utils/api.rst
namitdhameja: Its one or the other except for a very specific restarter case where Russell had asked for both. It would beat the point if nvrx logs always show up in both places .. The restarted case was as wanted the app to know.

rhewett-nv (approved):

apaithankar (approved):

PR #161, opened 2025-08-30
fix logger UT tests

namitdhameja:
Misc fixes:
1. Cursor introduces some bugs (wrong check for key workload_global_rank), fixing same.  
2. Removed cursor checks related to file rotation, that only happens for approaches where are truncating same file to 0 when max size reached.  
3. Remove the 8 processes logger cases to reduce the UT time from ~400 secs.
4. Made chronological logging flexible since there are cases where a write may appear on disk much later, this might still bite us since is not a bounded problem.

### DIFF
--- a/tests/shared_utils/test_logger.py
+++ b/tests/shared_utils/test_logger.py
@@ -32,10 +32,10 @@ Unit Test Usage:
 To run logger unit tests:
     PYTHONPATH=./src:$PYTHONPATH python -m unittest tests.shared_utils.test_logger
 
-To enable tests inclusive of 8 GPU's (e.g. H100), set RUN_EIGHT_PROC_TESTS=1:
-    PYTHONPATH=./src:$PYTHONPATH RUN_EIGHT_PROC_TESTS=1 python -m unittest tests.shared_utils.test_logger
+To enable tests inclusive of 8 GPU's (e.g. H100), set RUN_EXHAUSTIVE_TESTS=1:
+    PYTHONPATH=./src:$PYTHONPATH RUN_EXHAUSTIVE_TESTS=1 python -m unittest tests.shared_utils.test_logger
 
-The RUN_EIGHT_PROC_TESTS almost triples the UT time and therefore skipped by default.
+The RUN_EXHAUSTIVE_TESTS has 4x the UT time and therefore skipped by default.
 '''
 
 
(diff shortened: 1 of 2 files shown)


tests/shared_utils/test_logger.py
```diff
@@ -246,23 +246,29 @@ def test_four_proc(self):
         # gb200 has 4 GPU's, check that config
         self.multiple_processes(4, 2000, 1024)
 
+    '''
```
hexinw-nvidia: We can remove the commented tests if they are not required.

tests/shared_utils/test_logger.py
```diff
@@ -86,6 +86,9 @@ def check_a_file(self, filename, num_lines, global_id, local_id, chrono_on):
         line_count = 0
         curr_ts = 0
         curr_dt = 0
+        chrono_total = 0
+        chrono_success = 0
+        chrono_percent = 0.7
```
hexinw-nvidia, line 108: I am not very clear on the chrono_percent test. It doesn't look like the value is changed. 
We can wait till all logs are flushed and ensure all logs are in chronological order. Right?

tests/shared_utils/test_logger.py
namitdhameja: The 4 GPU's are for the GB200 case and the 8 GPU for H100. I would rather keep 8 GPUs in but it adds 200 sec to UT. Its still nice to have in UT. Let me restructure it a bit as having it in comments is not nice.

tests/shared_utils/test_logger.py
namitdhameja: The end-to-end flush time can range from milliseconds to a few seconds and depends on factors such as the OS load and disk setup. A Python 'flush()' pushes data from the application buffer to the OS, but the OS has its own cache, and flushing from the OS cache to the disk is a separate process. We can use 'os.fsync()' within a unit test flag (which I would rather avoid), though in a real workload, it is generally better for disk performance to let the driver manage this.

This review checks that 70% of log timestamps are in order ('chrono_percent = 0.7').

hexinw-nvidia: It should not. The reader should be able to see the whole data once "flush()" is committed to the OS, as long as the reader is referencing the same file. The data can come from the OS page cache even if it is not yet written to the disk.

tests/shared_utils/test_logger.py
hexinw-nvidia: Can we shorten the test duration? 200 sec in UT seems long.

tests/shared_utils/test_logger.py
namitdhameja: The aggregator reads each file buffer, sorts the contents, and writes the results to disk. If any file buffer runs empty, it re-reads that buffer to refresh its contents and resumes processing. However, race conditions can occur:

 - Process Wx generates a log message with a new timestamp, but it may be preempted at any point before or during the buffer write (user or kernel mode pre-emption). Meanwhile, Process Wy logs its own message later, yet its buffer is flushed first. As a result, the aggregator may miss Wx's entry in that cycle due to a delayed write or due to sequential file reading.

 - Process Wx writes to FileA and proceeds to create FileB. The aggregator, may not detect the existence of FileB in that cycle, and will only process files currently visible.

Overall, this approach generally performs reliably; in my environment, issues arise only about one in ten times. Therefore have made the corresponding unit test less stringent. I do not think further effort is needed to perfect timing-based behavior at this time, but I am open to suggestions.

tests/shared_utils/test_logger.py
namitdhameja: Its taking 80 sec now after the change to not run 8 processes by default; let me know if you think thats still too high 
-------
Ran 13 tests in 79.809s

tests/shared_utils/test_logger.py
hexinw-nvidia: I see your point now. I think we should focus on testing what’s deterministic. For example, have both Process Wx and Wy log to their own temp files, then let the log aggregator handle the aggregation. If we run the aggregator only after both Wx and Wy have finished logging, we should expect all logs to be captured in the final per-node file.

If the log aggregator can’t guarantee or quantify the success rate of sorting, then exposing a “chronological ordering percentage” setting seems confusing — and possibly flaky — for users.

namitdhameja: I think waiting for a file to be complete is not a good experience since the user could be tail logging on the aggregated file, but then will have to wait till the complete file is done and that actually may not happen till end of workload.
Agreed on chrono, it will generically work but by design its not a guaranteed. My thought is to release note this as its a good way to know if there is an actual interest in using this feature, if there isn't then there is no requirement to make it work for said race conditions.

namitdhameja:
Squash merged https://github.com/NVIDIA/nvidia-resiliency-ext/pull/168

PR #162, opened 2025-09-01
InJob logger regression

hexinw-nvidia:


src/nvidia_resiliency_ext/fault_tolerance/rank_monitor_server.py
```diff
@@ -108,11 +108,6 @@ def log_restarter_event(self, message, *args, **kwargs):
         if self.is_restarter_logger:
             self.log_for_restarter(message, *args, **kwargs)
 
-        # Get the nvrx logger directly
```
namitdhameja, line 111: there was a comment from @rhewett-nv to have the restart logging in both nvrx and with the application

rhewett-nv: We had intended to have this state also go to the NVRx logger.  But when both are going to `stderr` it is redundant.  It makes sense to go to both places when the NVRx logger is going to not stderr/out.

namitdhameja: right good point

namitdhameja (approved):

PR #165, opened 2025-09-03
Change the health check result logging at FR dump to `debug`

sbak5:
the health check logging in the FR dump is optional for debugging. 
The purpose of calling the routine is to covey health check results in FR dump trace for further analysis.


sbak5:
@hexinw-nvidia can you review this? this is a minor fix to make some printing in FR dump silent.

PR #167, opened 2025-09-03
Simulate health check failure on v0.5

hexinw-nvidia:
.


src/nvidia_resiliency_ext/fault_tolerance/launcher.py
```diff
@@ -1836,6 +1836,38 @@ def get_args_parser() -> ArgumentParser:
         "Template path to check if a NIC link is down.",
     )
 
+    parser.add_argument(
+        "--ft-simulate-failure-type",
+        "--ft_simulate_failure_type",
```
rhewett-nv, line 1841: no need for the underscore versions.  the other commands have them for bwd compatibility..

PR #169, opened 2025-09-03
Allow multiple AsyncCallsQueue in a process to enable different behav…

aartibasant:
QA identified a test scenario where within a single training loop, users may want to save multiple CP formats like Distributed or non Distributed CP.
Due to different collectives for dist or non-dist CP, having a singleton AsyncCallsQueue may cause hangs across different ranks.

Hence this PR eliminates the requirement for AsyncCallsQueue to be a Singleton. We now track all instances of AsyncCallsQueue using an ObjectTracker to cleanly abort all async wokers if a user wants to abort CP operations.


src/nvidia_resiliency_ext/checkpointing/async_ckpt/core.py
```diff
@@ -664,8 +665,7 @@ def __del__(self):
 
 def abort_nvrx_checkpoint():
     """Abort NVRx Checkpoint Utility. This will close the AsyncCallsQueue that manages async checkpoints"""
-    # we have a singleton persistent worker in our async calls queue
     # close the async calls queue which will ensure a clean restart
     # of the CP async process in subsequent async save requests.
-    async_queue_singleton = AsyncCallsQueue(persistent=True)
-    async_queue_singleton.close(abort=True)
+    for async_queue in AsyncCallsQueue.get_instances():
```
rhewett-nv, line 673: Adding this comment because I had to go down a path to understand why this is safe:

`.get_instances()` takes a set of weak references and makes strong references (the list above), so the queues cannot be GCed until this list is out of scope.  `WeakSet` will automatically remove GCed elements, so we should not have any `None` queues hitting this loop.

rhewett-nv (dismissed):

tests/checkpointing/unit/test_async_writer.py
```diff
@@ -219,3 +228,80 @@ def test_cached_metadata(self, tmp_path_dist_ckpt, async_queue):
...
+                ), "After abort async process must stop"
+
+            assert async_queue_dist is not None, "We expect a valid state of AsyncCallsQueue"
+            async_process_dist = async_queue_dist._get_async_caller()
+            if async_process_dist is not None:
+                assert (
+                    async_process_dist._debug_is_async_process_running() is False
+                ), "After abort async process must stop"
+
+            # Perform async saves for both dist CP and non-dict CP use cases.
```
diegs: nit: `non-dist`

tests/checkpointing/unit/test_async_writer.py
```diff
@@ -219,3 +228,80 @@ def test_cached_metadata(self, tmp_path_dist_ckpt, async_queue):
...
+            assert (
+                async_calls_queue_no_dist is not None
+            ), "We expect a valid state of AsyncCallsQueue"
+            async_process_no_dist = async_calls_queue_no_dist._get_async_caller()
+            if async_process_no_dist is not None:
+                assert (
+                    async_process_no_dist._debug_is_async_process_running() is False
+                ), "After abort async process must stop"
+
+            assert async_queue_dist is not None, "We expect a valid state of AsyncCallsQueue"
```
diegs: I am confused by these assertions. How could `async_queue_dist` possibly be `None` here? How does asserting that it is not `None` verify that it is in a valid state?

src/nvidia_resiliency_ext/checkpointing/async_ckpt/core.py
```diff
@@ -130,14 +131,18 @@ def execute_finalize_fns(self, validate_matching_call_idx: bool = True) -> int:
...
+class ObjectTracker(type):
+    def __init__(cls, name, bases, attrs):
+        super().__init__(name, bases, attrs)
+        cls._instances = weakref.WeakSet()
 
     def __call__(cls, *args, **kwargs):
-        if cls not in cls._instances:
-            cls._instances[cls] = super(Singleton, cls).__call__(*args, **kwargs)
-        return cls._instances[cls]
+        instance = super().__call__(*args, **kwargs)
```
diegs, line 140: Now that this is no longer a singleton, does this still achieve your original goal of ensuring duplicate instances are not created during in-process restarts?

diegs (changes requested):

src/nvidia_resiliency_ext/checkpointing/async_ckpt/core.py
aartibasant: The goal was never that duplicate instances should not be created during in-process restarts.
The goal was cleanup of AsyncCallsQueue state during inprocess restart.

Our original implementation with singletons was to esnure we had only 1 AsyncCallsQueue to clean up at any abort operation.

Now that we learnt that there are valid use cases that suggest the need for multiple AsyncCallsQueue, this PR is tracking all the instance creations and will clean up for all AsyncCallsQueue when abort is called.

tests/checkpointing/unit/test_async_writer.py
aartibasant: During the abort, we do not delete the AsyncCallsQueue object. We only kill the "process" that is spawned by the object. This check is validating that the `AsyncCallsQueue` object exists and that the process within it is not running

diegs: But how could `async_queue_dist` ever be `None` at this point? You instantiated it on line 240. I may be missing a python nuance but I don't see how this assertion could ever fail.

tests/checkpointing/unit/test_async_writer.py
```diff
@@ -219,3 +228,80 @@ def test_cached_metadata(self, tmp_path_dist_ckpt, async_queue):
...
+            async_process_no_dist = async_calls_queue_no_dist._get_async_caller()
+            # for the non_dist CP use case, only rank-0 is expected to trigger an async process
+            if async_process_no_dist is not None and torch.distributed.get_rank() == 0:
+                assert (
+                    async_process_no_dist._debug_is_async_process_running() is True
+                ), "After resume, we expect async process to be running on rank 0 for non dist async save"
+
+            assert async_queue_dist is not None, "We expect a valid state of AsyncCallsQueue"
+            async_process_dist = async_queue_dist._get_async_caller()
+            if async_process_dist is not None:
```
diegs: might be more complete and succinct to combine into:

```
assert async_process_dist is None or not async_process_dist._debug_is_async_process_running()
```

tests/checkpointing/unit/test_async_writer.py
aartibasant: I made this check because I call `abort_nvrx_checkpoint()` in line 260.
I agree we should never expect it to be None. It is a safeguard to avoid any faulty accesses in line 274.

tests/checkpointing/unit/test_async_writer.py
aartibasant: I think having separate assertions allow for clarity on what the failure is in the context of a unit test

tests/checkpointing/unit/test_async_writer.py
diegs: Can `abort_nvrx_checkpoint()` cause `async_queue_dist` to be `None`? That just doesn't seem possible to me. But again, if I am missing a python nuance please explain. Otherwise, I strongly prefer not to include asserts that are not actually asserting something, they obfuscate what is actually being tested. Again, if I am wrong about how it could ever become `None` then I would appreciate help understanding why.

tests/checkpointing/unit/test_async_writer.py
diegs: Ok, but in the first case you are only doing an `if`, which doesn't assert anything. The if statement implicitly says that another expected state is for `async_process_dist` to be `None`. If so, explicitly assert it.

tests/checkpointing/unit/test_async_writer.py
aartibasant: No `abort_nvrx_checkpoint()` is not expected to make this object None and the assertion is "not None".
I am happy to remove this assertion because in line 274, we access this object and if for any buggy reasons or future changes this object does become None, line 274 will fail and catch it anyways.

tests/checkpointing/unit/test_async_writer.py
aartibasant: You are looking at an outdated version. I have addressed this and I am explicitly asserting in the last commit.

diegs: Thanks, GitHub review UI is terrible.

diegs (approved):

PR #172, opened 2025-09-04
Flight recorder

sbak5:
Flight Recoder Analyzer, instantiated with  the base attribution module #141.

1. Parse per-rank flight recoder traces
2. entry matching to find out missing ranks per process group, 
3. Graph analysis considering the given process group ordering (e.g. intra-layer -> reduction -> default process group)

The outcome is the list of ranks to exclude.


src/nvidia_resiliency_ext/attribution/trace_analyzer/fr_attribution.py
```diff
@@ -0,0 +1,987 @@
...
+import sys
+from collections import Counter, defaultdict
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Dict, List, Tuple, Union
+
+from nvidia_resiliency_ext.attribution.base import NVRxAttribution
+from nvidia_resiliency_ext.attribution.utils import capture_stdout
+
+logging.basicConfig(level=logging.INFO)
```
hexinw-nvidia, line 17: Should we use NVRx logger? 
There is a mixed usage of "print" and logging in this file. What is the reason to use "print" but not logging?

src/nvidia_resiliency_ext/attribution/trace_analyzer/fr_attribution.py
```diff
@@ -0,0 +1,987 @@
...
+        """
+        Initialize the CollectiveAnalyzer class.
+
+        Args:
+            args: argparse.Namespace object containing the command line arguments
+
+        """
+        # the data structures to store the collective operations
+        self.collectives_by_file: Dict[str, List[Collective]] = {}
+        # the data structure to store the process group status per rank
```
hexinw-nvidia, line 92: I think an object-oriented design would help here. If we model CollectiveOp as a first-class object, then all we need is a key-value (SQL or NoSQL) database to store these CollectiveOp objects. We can explicitly define what constitutes the unique key (compound or single) to identify each CollectiveOp, and specify which fields should be indexed for fast lookup.

We could start with an in-memory database implementation, but define clean APIs around it. This way, migrating the data layer to a real SQL or NoSQL backend in the future would be straightforward.

sbak5: I think using any in-memory database is overkill. It simply corresponds to each value in the dictionary because the input is pickled dictionary of the data in PyTorch. This class corresponds exactly to the structure of the dictionary.

src/nvidia_resiliency_ext/attribution/trace_analyzer/fr_attribution.py
sbak5: the module aggregates string from the parsing and uses attr module with the string so using print for that to bypass  string buffer.

PR #183, opened 2025-09-11
Trace collector to use nvrx logs

namitdhameja:
1. https://nvbugspro.nvidia.com/bug/5497923 : [In-process][FR Trace] FR trace logs are incomplete when distributed logging is enabled

2. The backtrace logger support works but is incomplete, the complete fix is to append to previous line when the expected log format is not found.


namitdhameja:
> Is committing to the 0.5 branch still required?

The bug is opened on the rel branch, in the interest of closing it as I take it that QA won't test the feature on main branch. I would rather not have the feature wait for 0.6 to test. The fact that the feature has no release vehicle is an issue, but to me orthogonal.

rhewett-nv (approved):

PR #184, opened 2025-09-11
NVRX Logger: backtrace indentation fix and trace collector to use nvrx 

namitdhameja:
Multiple Double commits:

https://nvbugspro.nvidia.com/bug/5497923 : [In-process][FR Trace] FR trace logs are incomplete when distributed logging is enabled

The nvrx logger wasn't preserving backtrace indentation.

Double logging from RankMonitorLogger

### DIFF
--- a/src/nvidia_resiliency_ext/shared_utils/log_node_local_tmp.py
+++ b/src/nvidia_resiliency_ext/shared_utils/log_node_local_tmp.py
@@ -154,8 +154,10 @@ class LogMessage:
     def __init__(self, log_message: str):
         self.log_message = log_message
         self.hash_table = {}
+        self.log_message_valid = False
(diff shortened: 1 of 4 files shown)


src/nvidia_resiliency_ext/shared_utils/log_node_local_tmp.py
```diff
@@ -367,12 +369,21 @@ def _process_message_file(self, msg_file: str):
 
         # Process each line
```
hexinw-nvidia, line 370: Maybe worth adding some comments to help people to understand the following log message processing logic.

namitdhameja: Have added same

hexinw-nvidia (approved):

PR #185, opened 2025-09-14
feat: Add Injob time profiling metrics

hexinw-nvidia:
feat: Add Injob time profiling metrics
    
    - Add new profiling.py module with PyTorch metrics integration
    - Implement ProfilingEvent enum for tracking key InJob events:
      * FAILURE_DETECTED, WORKER_TERMINATED
      * RENDEZVOUS_STARTED, RENDEZVOUS_COMPLETED
      * WORKER_START_STARTED, WORKER_START_COMPLETED
    - Add FaultToleranceProfiler class with thread-safe event recording
    - Integrate profiling events throughout fault tolerance lifecycle:
      * Record rendezvous timing in FtRendezvousHandler
      * Track worker lifecycle events in LocalElasticAgent
      * Log comprehensive timing metrics summary on shutdown
    - Calculate and report timing metrics by restart cycles:
      * Failure to termination time
      * Rendezvous duration
      * Worker start time
      * Total cycle time (startup vs restart)


src/nvidia_resiliency_ext/fault_tolerance/profiling.py
```diff
@@ -0,0 +1,354 @@
...
+        self._logger.info("=== Fault Tolerance Profiling Summary ===")
+        for event_type, event_measurements in by_event.items():
+            self._logger.info(f"{event_type}: {len(event_measurements)} events recorded")
+            for measurement in event_measurements:
+                utc_time = self._timestamp_to_utc_datetime(measurement.timestamp)
+                # Get cycle number for this event
+                event_key = f"{measurement.event.value}_{measurement.timestamp}_{measurement.node_id}_{measurement.rank}"
+                cycle_num = event_to_cycle.get(event_key, "Unknown")
+                self._logger.info(
+                    f"  - Cycle: {cycle_num}, Node: {measurement.node_id}, Rank: {measurement.rank}, "
```
rhewett-nv: For post-hoc analysis it will be useful to include the event_type in each entry

src/nvidia_resiliency_ext/shared_utils/profiling.py
```diff
@@ -0,0 +1,414 @@
...
+from datetime import datetime, timezone
+from enum import Enum
+from typing import Any, Dict, List, Optional
+
+from torch.distributed.elastic.metrics.api import put_metric
+
+from ..shared_utils.log_manager import LogConfig
+
+
+class ProfilingEvent(Enum):
```
apaithankar, line 31: Any reason we are creating a custom profiler and not using something standard or OneLogger?

hexinw-nvidia: We are using the PyTorch metrics system for the profiling. The NV OneLogger seems to have duplicate functions as the PyTorch metrics. There is no particular reason that we go with PyTorch metrics. Mainly, NVRx is tightly integrated with PyTorch.

apaithankar: Since Megatron-LM and Nemo are both using NV OneLogger and it seems it is better integrated into other services (log/event/metrics collecting), would using NV OneLogger not make it easier to integrate when we have to create a dashboard in FACT?

Not opposed to using TorchMetrics, just looking at the ecosystem for integration

src/nvidia_resiliency_ext/shared_utils/profiling.py
```diff
@@ -0,0 +1,117 @@
...
+
+# This file adds time profiling capabilities using nv one logger
+
+import logging
+import time
+from datetime import datetime, timezone
+from enum import Enum
+from typing import Optional
+
+from nv_one_logger.api.one_logger_provider import OneLoggerProvider
```
PytLab, line 24: @hexinw-nvidia does it mean you are relying on application for instantiation and configuration?

src/nvidia_resiliency_ext/shared_utils/profiling.py
```diff
@@ -0,0 +1,117 @@
...
+                attributes.add("timestamp_ms", int(timestamp * 1000))
+                attributes.add("cycle", self._current_cycle)
+                if node_id:
+                    attributes.add("node_id", node_id)
+                if rank is not None:
+                    attributes.add("rank", rank)
+
+                # Create and record the event
+                event_obj = Event.create(f"ft.{event.value}", attributes)
+                OneLoggerProvider.instance().recorder.event(None, event_obj)
```
PytLab, line 76: @hexinw-nvidia Note that in nv-one-logger API design, event need to belong to one span e.g., application span or any custom span. the `Recorder.start` can create one and recorder will record the start and end timestamps of it.

PR #196, opened 2025-10-05
feat: Add infrastructure rank support and optimize section monitoring

hexinw-nvidia:
1. Infrastructure Group Rank Assignment
       - Add `use_infra_group_rank` config option to use
         infrastructure-provided ranks (e.g., SLURM_PROCID) instead of
         sorted participant-based assignment
    
    2. Selective Section Monitoring
       - Only send IPC messages for sections configured in
         `rank_section_timeouts`
       - Sections not in config have near-zero overhead (no IPC, no timing)
    
    3. Unidirectional Section Communication
       - Add `skip_section_response` config option (default: True)
       - Section and heartbeat messages sent without waiting for server
         response
       - Significantly reduces latency for high-frequency monitoring
         operations

    4. Allow null values for timeout CLI arguments
    
        Change timeout arguments from type=float to type=str to support
       'null'/'none'/'' values for disabling timeouts.

### DIFF
--- a/src/nvidia_resiliency_ext/fault_tolerance/_ft_rendezvous.py
+++ b/src/nvidia_resiliency_ext/fault_tolerance/_ft_rendezvous.py
@@ -233,6 +233,10 @@ class RendezvousSettings:
             If set to True (default), nodes from the redundancy list and new arrivals are migrated
             to the wait list. If set to False, new arrivals will be moved to the redundancy list
             and will wait there until the next rendezvous round.
+        use_infra_group_rank:
+            Whether to use infrastructure group rank for rank assignment instead of sorted
+            participant-based assignment. If True, ranks are read from SLURM_PROCID (in SLURM
+            environments) or GROUP_RANK (set by launcher) environment variables.
     """
 
     run_id: str
@@ -242,6 +246,7 @@ class RendezvousSettings:
     keep_alive_interval: timedelta
     keep_alive_max_attempt: int
     upscaling_enabled: bool = True
+    use_infra_group_rank: bool = False
 
 
 @dataclass(eq=True, order=True, frozen=True)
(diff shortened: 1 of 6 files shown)


src/nvidia_resiliency_ext/fault_tolerance/_ft_rendezvous.py
```diff
@@ -242,6 +246,7 @@ class RendezvousSettings:
     keep_alive_interval: timedelta
     keep_alive_max_attempt: int
     upscaling_enabled: bool = True
+    use_infra_group_rank: bool = False
```
rhewett-nv: We should default to the infra's assignment

hexinw-nvidia: Done

rhewett-nv (dismissed):

src/nvidia_resiliency_ext/fault_tolerance/_ft_rendezvous.py
```diff
@@ -789,8 +794,21 @@ def _add_to_participants(self) -> None:
...
+        if self._settings.use_infra_group_rank:
+            # Try SLURM_PROCID first (set by SLURM), then fall back to GROUP_RANK (set by launcher)
+            infra_rank_str = os.getenv('SLURM_PROCID', os.getenv('GROUP_RANK', '-1'))
+            infra_rank = int(infra_rank_str)
+            if infra_rank < 0:
+                log.warning(
+                    f"use_infra_group_rank is enabled but neither SLURM_PROCID nor GROUP_RANK "
+                    f"env var is set or valid. Node {self._node} will use placeholder rank."
+                )
+            state.participants[self._node] = infra_rank
```
apaithankar, line 809: Why is `state.participants[self._node]` set to -1 when SLURM_PROCID or GROUP_RANK is not set, but set to 0 when `use_infra_group_rank` is disabled?
Are they handled differently based on 0 or -1 values?

apaithankar: It says use placeholder 0, but I see -1 being set.

hexinw-nvidia: Yes, there is. When "use_infra_group_rank" is True, we expect the group_rank to be valid from the infra level. 
A warning (line 804) is given to this participant if it is not valid. Eventually, if such a participant exists in the final _assign_ranks() call, it will be a ValueError exception to the whole Rendezvous.

src/nvidia_resiliency_ext/fault_tolerance/_ft_rendezvous.py
```diff
@@ -874,16 +892,61 @@ def _remove_from_redundancy_list(self) -> None:
...
+        """
+        Assign ranks to participants in the rendezvous.
+
+        Behavior depends on use_infra_group_rank and previous assignments:
+
+        1. If use_infra_group_rank=True AND prev is empty (first rendezvous):
+           - Use infrastructure ranks directly from SLURM_PROCID or GROUP_RANK
+           - Validates that all ranks are in range [0, world_size) and unique
+
+        2. If prev is not empty (subsequent rendezvous, including after failures):
```
apaithankar, line 908: This needs to be done at the node level and not at individual rank level. So "rank blocks" will have to be removed and filled.

src/nvidia_resiliency_ext/fault_tolerance/_ft_rendezvous.py
hexinw-nvidia: Placeholder comment is fixed. Thanks for the catch.

PR #289, opened 2026-04-03
chat_open_ai

helisha91:


pyproject.toml
namitdhameja, line 1: is this needed now, asking since have moved to openai?

src/nvidia_resiliency_ext/attribution/log_analyzer/posting_markdown.py
```diff
@@ -30,12 +31,15 @@ def format_attribution_markdown(
     jid = job_id if job_id else "unknown"
     attr = attribution_text if attribution_text else "No attribution available"
     expl = auto_resume_explanation if auto_resume_explanation else "No explanation available"
+    lp = log_path if log_path else "Log path not found"
```
namitdhameja: won't empty give the same info anyways ie path not found

src/nvidia_resiliency_ext/attribution/log_analyzer/config.py
```diff
@@ -47,7 +47,8 @@ class LogSageExecutionConfig:
     use_lib_log_analysis: bool = False
     #: Subprocess MCP server (:func:`~nvidia_resiliency_ext.attribution.mcp_integration.mcp_client.get_server_command`).
     mcp_server_log_level: str = "INFO"
-    llm_model: str = "nvdev/nvidia/llama-3.3-nemotron-super-49b-v1"
+    llm_model: str = "nvidia/qwen/qwen-235b"
```
namitdhameja: consider having a single place at least in log_analyzer for model and base. We are having the same variables redefined multiple times in this analyzer. A single default is better.

.idea/inspectionProfiles/profiles_settings.xml
namitdhameja, line 1: did you mean to add in the .dir?

namitdhameja (changes requested):

pyproject.toml
helisha91: @namitdhameja 
No need, will remove

src/nvidia_resiliency_ext/attribution/log_analyzer/config.py
helisha91: So, we should create consts?

.idea/inspectionProfiles/profiles_settings.xml
helisha91: No... I wil remove

src/nvidia_resiliency_ext/attribution/log_analyzer/posting_markdown.py
helisha91: @namitdhameja 
What do you mean?

src/nvidia_resiliency_ext/attribution/log_analyzer/config.py
namitdhameja: I guess there is duplication in LogSageExecutionConfig and LogAnalyzerConfig. We could refactor to remove one. Or just have say DEFAULT_LLM_MODEL et all in both.

src/nvidia_resiliency_ext/attribution/log_analyzer/posting_markdown.py
namitdhameja: I am suggesting that if the log path is not there then just have it empty. Its easier to parse empty values instead of having an explicit message, which then you have to parse for validity.

namitdhameja (changes requested):

src/nvidia_resiliency_ext/attribution/log_analyzer/config.py
helisha91: Add to: services/nvrx_attrsvc/config.py

# Service-specific constants
DEFAULT_HOST = "0.0.0.0"
DEFAULT_PORT = 8000
PRINT_PREVIEW_MAX_BYTES = 4096  # Max bytes to return for /print endpoint
DEFAULT_LLM_MODEL = "nvidia/qwen/qwen-235b"
DEFAULT_LLM_BASE_URL = "https://inference-api.nvidia.com/v1"

namitdhameja (dismissed):

namitdhameja (dismissed):

namitdhameja (dismissed):

namitdhameja (dismissed):

namitdhameja (dismissed):

namitdhameja (approved):

PR #299, opened 2026-04-17
Fix: out-of-section timeout never fires for section-less workloads

hexinw-nvidia:
Add self.cfg.rank_section_timeouts check. Activate the warmup guard when section timeouts are actually configured.


hexinw-nvidia:
@greptile

sbak5 (approved):

PR #302, opened 2026-04-21
shared_utils: add NVLinkWindowHealthCheck for replay/recovery counter monitoring

ashrot-ops:
## Summary

Adds `NVLinkWindowHealthCheck` to `src/nvidia_resiliency_ext/shared_utils/health_check.py`.

The existing `NVLHealthCheck` detects disabled NVLink ports, but does not track transient replay/recovery errors that indicate a degrading link before it goes down. `NVLinkWindowHealthCheck` fills this gap by sampling error counters every interval and reporting failure only when errors are persistent across multiple windows — reducing false positives from single transient events.

## What changed

- New class `NVLinkWindowHealthCheck(PynvmlMixin)` in `shared_utils/health_check.py`
- Two new imports: `deque` from `collections`, `Tuple` from `typing`

## How it works

Each call to `_perform_health_check()` is one **window**. A window is flagged when any NVLink port shows a delta > 0 in `DL_REPLAY` or `DL_RECOVERY` error counters vs. the previous call. Returns `False` when any port exceeds `threshold` flagged windows within the last `window_count` calls.

### Parameters

| Parameter | Default | Description |
|---|---|---|
| `device_index` | `None` | GPU index to monitor; `None` checks all GPUs |
| `interval` | `60` | Seconds between async health checks |
| `window_count` | `30` | Rolling window size (number of past checks to consider) |
| `threshold` | `15` | Max flagged windows allowed before reporting failure |
| `on_failure` | `None` | Callback invoked on failure |

### Counter reading — three tiers (first that works wins per port)

**Tier 1 — GB200 / newer drivers (preferred)**
`nvmlDeviceGetFieldValues` with per-port field IDs and `scopeId = port`:
- `NVML_FI_DEV_NVLINK_ERROR_DL_REPLAY` (161)
- `NVML_FI_DEV_NVLINK_ERROR_DL_RECOVERY` (162)

**Tier 2 — older hardware (ports 0–5 only)**
`nvmlDeviceGetFieldValues` with per-lane field IDs L0–L5:
- `NVML_FI_DEV_NVLINK_REPLAY_ERROR_COUNT_L{n}` (46–51)
- `NVML_FI_DEV_NVLINK_RECOVERY_ERROR_COUNT_L{n}` (53–58)

**Fallback — legacy (drivers ≤ 450)**
`nvmlDeviceGetNvLinkErrorCounter` with `NVML_NVLINK_ERROR_DL_REPLAY` / `NVML_NVLINK_ERROR_DL_RECOVERY`. May return `NVML_ERROR_NOT_SUPPORTED` on GB200.

A one-time warning is logged if GB200 tier-1 field IDs exist in pynvml but the driver returns not-supported, and again if all APIs fail on a port (health check skipped for that port).

## What to review

- Default values for `window_count=30` and `threshold=15` — are these appropriate for the expected polling interval?
- The three-tier API fallback logic — especially the `_tier1_api_supported` / `_tier2_api_supported` flag tracking
- Whether `NVLinkWindowHealthCheck` should be exported from `shared_utils/__init__.py`

### DIFF
--- a/src/nvidia_resiliency_ext/shared_utils/health_check.py
+++ b/src/nvidia_resiliency_ext/shared_utils/health_check.py
@@ -1164,8 +1164,12 @@ class NVLinkWindowHealthCheck(PynvmlMixin):
 
         # Tier 2: per-lane field IDs L0-L5 (older hardware, ports 0-5 only).
         if self._tier2_api_supported is not False and port < 6:
-            replay_lane = getattr(self.pynvml, f"NVML_FI_DEV_NVLINK_REPLAY_ERROR_COUNT_L{port}", None)
-            recovery_lane = getattr(self.pynvml, f"NVML_FI_DEV_NVLINK_RECOVERY_ERROR_COUNT_L{port}", None)
+            replay_lane = getattr(
+                self.pynvml, f"NVML_FI_DEV_NVLINK_REPLAY_ERROR_COUNT_L{port}", None
+            )
+            recovery_lane = getattr(
+                self.pynvml, f"NVML_FI_DEV_NVLINK_RECOVERY_ERROR_COUNT_L{port}", None
+            )
             field_ids = [fid for fid in (replay_lane, recovery_lane) if fid is not None]
             if field_ids:
                 try:
--- a/src/nvidia_resiliency_ext/shared_utils/health_check.py
+++ b/src/nvidia_resiliency_ext/shared_utils/health_check.py
@@ -24,9 +24,9 @@ import subprocess  # nosec B404
 import sys
 import threading
 import traceback
-from collections import defaultdict
+from collections import defaultdict, deque
 from functools import wraps
-from typing import Any, Callable, Dict, Optional, Union
+from typing import Any, Callable, Dict, Optional, Tuple, Union
 from urllib.parse import quote_plus
 
 import defusedxml.ElementTree as ET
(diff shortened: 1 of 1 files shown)


ashrot-ops:
@namitdhameja CI checks haven't triggered yet — could you approve the workflow run when you get a chance? Thanks!

hexinw-nvidia:
Are the Greptile reviews getting addressed?

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -971,6 +971,257 @@ def _check_nvl_links_for_device(self, device_index: int) -> bool:
...
+        self._prev_counters: Dict[Tuple[int, int], int] = {}
+        # Separate probed flags for tier 1 (GB200) and tier 2 (older HW).
+        # None = not yet probed; True = supported; False = not supported.
+        self._tier1_api_supported: Optional[bool] = None
+        self._tier1_not_supported_warned = False
+        self._tier2_api_supported: Optional[bool] = None
+        # Warn only once when all APIs are unsupported for a port.
+        self._api_not_supported_warned = False
+
+    async def async_check(self) -> None:
```
namitdhameja, line 1107: I don't see the callee code ie where is it being invoked. It could be that the callee is viewed as integration code, so a different PR will tie this into the ft_launcher health framework.

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -971,6 +971,257 @@ def _check_nvl_links_for_device(self, device_index: int) -> bool:
...
+        """Periodically checks NVLink event windows; invokes on_failure when threshold is exceeded."""
+        while True:
+            await asyncio.sleep(self.interval)
+            result = await self._check_health()
+            if not result and self.on_failure:
+                await self.on_failure()
+
+    async def _check_health(self) -> bool:
```
namitdhameja, line 1119: Same here, the checks can be called at the life stage ie process start and stop. Or periodically, and am not clear on how the integration is. I guess, the thought could be to call it periodically, but make the decision at life stage. In that case, would hook into both the periodic and the lie stage. Though you would need be careful as can invoke  _perform_health_check by two different threads at the same time, and might need a lock.

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -971,6 +971,257 @@ def _check_nvl_links_for_device(self, device_index: int) -> bool:
...
+    A one-time warning is logged if GB200 tier-1 field IDs exist in pynvml but the
+    driver returns not-supported, and again if all APIs fail on a port.
+    """
+
+    def __init__(
+        self,
+        device_index: Optional[int] = None,
+        interval: int = 60,
+        window_count: int = 30,
+        threshold: int = 15,
```
namitdhameja, line 1034: Is there a documentation on where the thresholds are coming from? ie say a PID. If its an external facing doc then a reference will help.

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -971,6 +971,257 @@ def _check_nvl_links_for_device(self, device_index: int) -> bool:
...
+        Tries tier 1 (GB200) then tier 2 (older HW) independently; each tier's
+        supported flag is tracked separately so a tier-2 failure never blocks tier 1.
+        Returns None if both tiers are unsupported.
+        """
+        nvml_success = getattr(self.pynvml, "NVML_SUCCESS", 0)
+
+        # Tier 1: GB200 per-port (scopeId = port number).
+        # Field IDs 161/162 with scopeId are supported on newer drivers (GB200+).
+        if self._tier1_api_supported is not False:
+            replay_fid = getattr(self.pynvml, "NVML_FI_DEV_NVLINK_ERROR_DL_REPLAY", None)
```
namitdhameja: https://docs.nvidia.com/deploy/nvml-api/group__nvmlFieldValueEnums.html#group__nvmlFieldValueEnums_1gb07c22b7347336b985c6ea096e64186f - This is unsupported for Blackwell+
the doc says unsupported but the code comments say supported for new drivers, so I am not sure of what counters are supported.

namitdhameja (changes requested):

src/nvidia_resiliency_ext/shared_utils/health_check.py
ashrot-ops: This is designed as a periodic-only check, similar to a flap counter — it accumulates a rolling 
 window of NVLink error deltas over time. A single life-stage call would be meaningless since    
 _prev_counters is empty on first call (no delta to compute). Thread safety is already handled by
 @with_pynvml_lock on _perform_health_check, which uses a global threading.RLock() consistent 
 with all other checks in this file.

src/nvidia_resiliency_ext/shared_utils/health_check.py
ashrot-ops: This follows the same pattern as the existing health check classes in this file (e.g.,  GPUHealthCheck, NicHealthCheck). The integration point (wiring async_check into the ft_launcher health framework) is out of scope for this PR and will be a separate ask

src/nvidia_resiliency_ext/shared_utils/health_check.py
ashrot-ops: The current reference is related to other counters that are not extracted. you can see 
https://nvbugspro.nvidia.com/bug/6061118?commentNumber=22

Latest FAIL criteria for NVLinks:
NVLink PHY check #1 : Fail link if  PLR RX Error Rate is 2% or higher for two consecutive 60-second samples. The PLR Tx Retransmission Rate is calculated as:
PLR Tx Retransmission Rate = blocks re-transmitted / total blocks transmitted
plr-xmit-retry-codes / plr-xmit-codes * 100%
NVLink PHY check #2:  Fail link if PLR TX Retransmission Rate is 2% or higher for two consecutive 60-second measurement intervals. 
PLR Rx Error Rate = received erroneous blocks / total blocks received
plr-rcv-codes-err / plr-rcv-code * 100% // GB NVL 1.3.5+
plr-rcv-codes-err / (plr-rcv-codes + plr-rcv-codes-err) * 100% // GB NVL 1.3.3
NVLink PHY check #3: Fail link if any 60-second sampling period contains more than 2 successful recoveries. 
Eff BER threshold of 1e-9

The best approach will be to set slightly lower thresholds for health check.
I used effective errors as indication to Retransmission

src/nvidia_resiliency_ext/shared_utils/health_check.py
ashrot-ops: Thank you for catching this. Looking at the NVML docs, you're right that 161/162 may not be  supported on Blackwell+. The health check requires at least errors and recovery counters to be meaningful. Could you help clarify which counter APIs are supported on GB200/Blackwell+ for  NVLink errors? That will let me update the tier logic accordingly.

src/nvidia_resiliency_ext/shared_utils/health_check.py
namitdhameja: You could still accumulate the counters over the lifetime of the job, but only make an actual decision at a lifecycle boundary, such as a restart point. The alternative is to treat a health-check failure as fatal immediately by raising an exception and terminating the training job.

Based on your earlier comment, that policy should live with the caller and that will be a seperate PR. This PR can provide the signal; the code using it can decide whether to “kill job and exclude node right now” OR “just record score and wait until restart for any exclusion”.

I’m also not fully sure whether this should be modeled as a binary health decision or as a health score. If it is a score, we could simply prefer the healthiest available nodes, but that may add more complexity than we want right now.

src/nvidia_resiliency_ext/shared_utils/health_check.py
namitdhameja: Thanks. If this was an externally available doc then could have added a reference to that in nvrx docs. I don't know what nvbug access policy is so we can skip adding it to the docs.

src/nvidia_resiliency_ext/shared_utils/health_check.py
namitdhameja: Point being, while we are not encoding policy in this PR. It will be good to have your thoughts on same.

ashrot-ops: I agree, health score sounds more correct

src/nvidia_resiliency_ext/shared_utils/health_check.py
ashrot-ops: I will check and update

src/nvidia_resiliency_ext/shared_utils/health_check.py
namitdhameja: These worked for me on my local setup:
#define [NVML_FI_DEV_NVLINK_COUNT_LINK_RECOVERY_EVENTS](https://docs.nvidia.com/deploy/nvml-api/group__nvmlFieldValueEnums.html#group__nvmlFieldValueEnums_1g40045d78d663f88a72a04c8bec118fbe) 215
Number of times link went from Up to recovery, irrespective of the result.
#define [NVML_FI_DEV_NVLINK_COUNT_LINK_RECOVERY_FAILED_EVENTS](https://docs.nvidia.com/deploy/nvml-api/group__nvmlFieldValueEnums.html#group__nvmlFieldValueEnums_1g5bbd3f3881a8ad660dda714faaba0a88) 214
Number of times link went from Up to recovery, failed and link was declared down.
#define [NVML_FI_DEV_NVLINK_COUNT_LINK_RECOVERY_SUCCESSFUL_EVENTS](https://docs.nvidia.com/deploy/nvml-api/group__nvmlFieldValueEnums.html#group__nvmlFieldValueEnums_1g6a46bbe5bdb96ca850be5c6a5c47dc41) 213
Number of times link went from Up to recovery, succeeded and link came back up.

https://docs.nvidia.com/deploy/nvml-api/group__nvmlFieldValueEnums.html

ashrot-ops: Thanks Namit — good catch. Pushed [752af26](https://github.com/ashrot-ops/nvidia-resiliency-ext/commit/752af26) which swaps the tier-1 field IDs from 161/162 (DL_REPLAY/DL_RECOVERY, which were unsupported on your setup and are documented as unsupported on Blackwell+) to three publicly-documented per-port indicators that you verified work:

- **215** `NVML_FI_DEV_NVLINK_COUNT_LINK_RECOVERY_EVENTS` — Up→recovery transitions (any outcome)
- **295** `NVML_FI_DEV_NVLINK_PLR_XMIT_RETRY_BLOCKS` — NVLink5 PHY-layer retry pressure
- **208** `NVML_FI_DEV_NVLINK_COUNT_RCV_ERRORS` — Rx packets with errors

All three are read in a single `nvmlDeviceGetFieldValues` call with `scopeId = port`, summed into the per-port total, and any delta > 0 between samples flags the current window. The existing 30-sample sliding window (`deque(maxlen=window_count)`) is unchanged, so the threshold semantics still answer 'how many of the last 30 windows showed an issue.' Tier-2 (per-lane 46–51 / 53–58) and the legacy fallback are kept for older hardware/drivers.

ashrot-ops:
Good catch — fixed via per-FID baselines (your Option 2). `_prev_counters`
is now `Dict[(device, port), Dict[FID, value]]` and
`_sample_nvlink_events` flags the window only when a counter present in
BOTH calls has a positive diff — new or missing FIDs can't generate a
phantom delta. First successful read appends `False` (baseline only);
failed reads don't advance the window; deque keeps the last 30 via
`maxlen=30`. Verified off-tree against your exact example + real
increment + per-port isolation + failed-read gap + baseline persistence
+ deque cap.

New commit: `7c7e797 shared_utils: track NVLink counters per-FID instead
of summed total`. Visible in the **Commits** tab and as a timeline event
in the **Conversation** tab; the **Files changed** tab shows the new
aggregate diff. Re-scan when ready.

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -971,6 +971,253 @@ def _check_nvl_links_for_device(self, device_index: int) -> bool:
...
+                    if not self._tier1_not_supported_warned:
+                        logger.warning(
+                            f"NVLinkWindowHealthCheck: GB200 field IDs (161/162) not supported "
+                            f"on device {device_id} port {port}. "
+                            f"Falling back to tier-2 / legacy API."
+                        )
+                        self._tier1_not_supported_warned = True
+                    self._tier1_api_supported = False
+                except self.pynvml.NVMLError:
+                    pass  # fall through to tier 2
```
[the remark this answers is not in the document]
ashrot-ops: Fixed in 566892b. The tier-1 `except NVMLError` handler now sets `_tier1_api_supported = False` on a first-time probe (guarded by `is None` to preserve the sticky-True latch). Persistent NVMLError on first probe disables tier-1 to avoid retrying every port every call; a transient NVMLError after the tier was already confirmed `True` no longer demotes it. Resolving.

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -971,6 +971,253 @@ def _check_nvl_links_for_device(self, device_index: int) -> bool:
...
+            handle = self.pynvml.nvmlDeviceGetHandleByIndex(device_id)
+            for port in range(max_links):
+                total = self._read_port_error_total(handle, device_id, port)
+                if total is None:
+                    continue
+                key = (device_id, port)
+                prev = self._prev_counters.get(key)
+                self._prev_counters[key] = total
+                if prev is not None:
+                    port_events[key] = total > prev
```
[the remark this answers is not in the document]
ashrot-ops: Marked outdated after the per-FID rewrite. The same comparison style is in the per-FID path, but for these specific counters (NVLink recovery events / retry blocks) at typical link rates, 2^32 wrap-time exceeds a training-job lifetime by orders of magnitude. Tracking as future hardening rather than blocking this PR. Resolving.

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -971,6 +971,253 @@ def _check_nvl_links_for_device(self, device_index: int) -> bool:
...
+                logger.warning(f"NVLinkWindowHealthCheck NVML shutdown error: {e}")
+
+        # Update per-port rolling window and check threshold.
+        all_healthy = True
+        for key, had_events in port_events.items():
+            if key not in self._window_history:
+                self._window_history[key] = deque(maxlen=self._window_count)
+            self._window_history[key].append(had_events)
+            flagged = sum(self._window_history[key])
+            if flagged > self.threshold:
```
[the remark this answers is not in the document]
ashrot-ops: Clarified the constructor docstring instead of changing semantics (566892b): `threshold` is now documented as "Maximum flagged windows still considered healthy at the per-port level. A port is bad when the number of flagged windows in its rolling deque exceeds this value (i.e. at `threshold + 1` or more)." The `>` comparison in code is intentional and matches that wording. Resolving.

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -971,6 +971,253 @@ def _check_nvl_links_for_device(self, device_index: int) -> bool:
...
+        # Warn only once when all APIs are unsupported for a port.
+        self._api_not_supported_warned = False
+
+    async def async_check(self) -> None:
+        """Periodically checks NVLink event windows; invokes on_failure when threshold is exceeded."""
+        while True:
+            await asyncio.sleep(self.interval)
+            result = await self._check_health()
+            if not result and self.on_failure:
+                await self.on_failure()
```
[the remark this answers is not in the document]
ashrot-ops: Documented the async-callable contract in the `on_failure` parameter docstring (566892b): "Must be an `async def` function — synchronous callables will raise `TypeError`. Matches the await pattern of `NVLHealthCheck.async_check`." Keeping the unconditional `await` for consistency with the existing `NVLHealthCheck` pattern. Resolving.

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -971,6 +971,257 @@ def _check_nvl_links_for_device(self, device_index: int) -> bool:
...
+                    successful = [r for r in results if r.nvmlReturn == nvml_success]
+                    if successful:
+                        self._tier2_api_supported = True
+                        return sum(r.value.uiVal for r in successful)
+                except self.pynvml.NVMLError as e:
+                    logger.warning(
+                        f"NVLinkWindowHealthCheck: nvmlDeviceGetFieldValues tier-2 not supported "
+                        f"on device {device_id}: {e}. Falling back to legacy API."
+                    )
+                self._tier2_api_supported = False
```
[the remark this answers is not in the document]
ashrot-ops: Fixed in 566892b. Both the empty-results path and the `except NVMLError` handler in tier-2 now guard `_tier2_api_supported = False` with `if self._tier2_api_supported is None:`, so a previously-confirmed tier won't regress when a later port returns empty success. Symmetric fix applied to tier-1 in the same commit. Resolving.

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -971,6 +971,267 @@ def _check_nvl_links_for_device(self, device_index: int) -> bool:
...
+            getattr(self.pynvml, "NVML_FI_DEV_NVLINK_COUNT_RCV_ERRORS", 208),
+        )
+        if self._tier1_api_supported is not False:
+            queries = [(fid, port) for fid in TIER1_FIDS]
+            try:
+                results = self.pynvml.nvmlDeviceGetFieldValues(handle, queries)
+                successful = [r for r in results if r.nvmlReturn == nvml_success]
+                if successful:
+                    self._tier1_api_supported = True
+                    return sum(r.value.uiVal for r in successful)
```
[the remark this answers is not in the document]
ashrot-ops: Addressed in commit a85238e (already on the PR before this round). `_prev_counters` is now `Dict[(device, port), Dict[FID, value]]` and `_sample_nvlink_events` only flags a window when a counter present in both the previous and current sample has a positive diff: `any(fid in prev and val > prev[fid] for fid, val in counters.items())`. New FIDs joining the baseline don't fire a phantom delta; FIDs that disappear one call don't either. Resolving.

src/nvidia_resiliency_ext/shared_utils/health_check.py
ashrot-ops: Going with skip per your suggestion — the source nvbug is internal-only so not suitable for the nvrx docs. Resolving.

ashrot-ops:
Yes — all 6 greptile findings addressed in the latest push (c809766):

- Tier-1 / tier-2 sticky-flag fixes (566892b)
- Per-FID baseline tracking (a85238e)
- Threshold and on_failure docstring clarifications (566892b)
- Locking already serialized via `@with_pynvml_lock`
- Counter wraparound deferred (greptile marked outdated; wrap-time exceeds job lifetime)

Also added a per-GPU rollup (c809766, `gpu_ports_threshold` parameter, default 1 preserves prior behavior) so the check's external framing matches the per-GPU decision pattern of NVLHealthCheck and the other health checks in this file. Branch rebased onto latest main and ready for review.

namitdhameja (dismissed):

namitdhameja (approved):

PR #307, opened 2026-04-22
Auto-select log aggregation tier.

hexinw-nvidia:
Default --ft-log-aggregator-count to 0 (auto): single root server for ≤1536 nodes, one leaf per 1536 nodes beyond that. Eliminates manual tuning for most jobs.

Increase per_leaf worker headroom from +10 to +100 to absorb load imbalance when SLURM_PROCID sharding groups multiple ranks per physical node onto the same leaf.


hexinw-nvidia:
@greptile

sbak5 (approved):

PR #318, opened 2026-04-30
feat: Added ft_launcher managed attrsvc

hexinw-nvidia:
Support ft_launcher to manage the attribution process life cycle.

Default the attribution service launcher path to disabled unless --ft-attrsvc-host is set, and require per-cycle application logs when the service is enabled. Add managed attrsvc lifecycle handling, startup validation, port-overlap checks, cleanup, and focused unit coverage.

Package nvrx_attrsvc and nvrx_smonsvc in the main NVRx wheel while keeping their service dependencies behind the existing attribution extra. Plain wheel installs remain dependency-clean; users opt into the service stack with nvidia_resiliency_ext-*.whl[attribution]. Add service entrypoint wrappers that report a clear install hint when optional dependencies are missing.

### DIFF
--- a/docs/source/fault_tolerance/usage_guide.rst
+++ b/docs/source/fault_tolerance/usage_guide.rst
@@ -112,30 +112,64 @@ Validation behavior:
 Attribution service integration
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Enable artifact analysis (e.g., logs) during rendezvous health checks by pointing to a running attribution service.
-The feature is enabled by specifying both host and port.
+Per-cycle application logs do not enable attribution by themselves. To enable attribution, set
+``--ft-attribution-endpoint``. The endpoint value ``localhost`` makes ``ft_launcher`` run the
+attribution service on the TCPStore host; other endpoints are treated as externally managed
+attribution services.
+External endpoints may use schemes such as ``http://``, ``grpc://``, or ``unix://``. The current
+in-job attribution client submits logs over HTTP(S); non-HTTP endpoint strings are preserved but do
+not add a new transport implementation.
+If ``--ft-attribution-endpoint`` is set, ``--ft-per-cycle-applog-prefix`` is required because the
+attribution service analyzes the per-cycle application logs.
+
+The service code is included in the NVRx wheel, but the service dependencies are optional.
+Install the wheel with the ``attribution`` extra before running a launcher-managed attribution
+service:
+
+.. code-block:: bash
+
+   python -m pip install 'nvidia_resiliency_ext-<version>-<tags>.whl[attribution]'
+
+Plain ``python -m pip install nvidia_resiliency_ext-*.whl`` does not install the attribution
+service dependencies.
 
 * CLI:
 
-  - ``--ft-attrsvc-host <HOST>`` (alias: ``--ft_attrsvc_host``)
-  - ``--ft-attrsvc-port <PORT>`` (alias: ``--ft_attrsvc_port``)
+  - ``--ft-attribution-endpoint <ENDPOINT>`` (alias: ``--ft_attribution_endpoint``), default disabled
+  - ``--ft-attribution-llm-api-key-file <PATH>`` (alias: ``--ft_attribution_llm_api_key_file``)
+  - ``--ft-attribution-llm-base-url <URL>`` (alias: ``--ft_attribution_llm_base_url``)
+  - ``--ft-attribution-llm-model <MODEL>`` (alias: ``--ft_attribution_llm_model``)
+  - ``--ft-attribution-startup-timeout <SECONDS>`` (alias: ``--ft_attribution_startup_timeout``), default ``20``
+
+  The managed attribution app-log directory is derived from
+  ``dirname(realpath(--ft-per-cycle-applog-prefix))``. Its stdout/stderr log is derived
+  from ``--ft-per-cycle-applog-prefix`` as ``*_attribution.log``. The managed service listens on
+  ``127.0.0.1:50050`` and is exposed to the in-job client as ``http://localhost:50050``.
+
+  The managed attribution API key must come from ``--ft-attribution-llm-api-key-file`` or inherited
+  ``LLM_API_KEY_FILE``. If neither points to a readable file, the TCPStore-host launcher fails
+  before starting the attribution service.
 
   Example:
 
   .. code-block:: bash
 
      ft_launcher \
-       --ft-attrsvc-host 127.0.0.1 \
-       --ft-attrsvc-port 8000 \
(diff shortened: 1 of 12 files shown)


hexinw-nvidia:
@greptile

docs/source/fault_tolerance/usage_guide.rst
```diff
@@ -112,30 +112,59 @@ Validation behavior:
...
+   python -m pip install 'nvidia_resiliency_ext-<version>-<tags>.whl[attribution]'
+
+Plain ``python -m pip install nvidia_resiliency_ext-*.whl`` does not install the attribution
+service dependencies.
 
 * CLI:
 
-  - ``--ft-attrsvc-host <HOST>`` (alias: ``--ft_attrsvc_host``)
-  - ``--ft-attrsvc-port <PORT>`` (alias: ``--ft_attrsvc_port``)
+  - ``--ft-attrsvc-host <HOST>`` (alias: ``--ft_attrsvc_host``), default disabled
```
namitdhameja: This ties attribution to a service framework, but there are variations such as MCP possible or simply running it as a library is also possible. Perhaps,
--ft-attribution-backend or --ft-attribution this would work for [mcp | lib | unix:// | http://]. For now we can only accept say <uds:file-name>

docs/source/fault_tolerance/usage_guide.rst
```diff
@@ -112,30 +112,59 @@ Validation behavior:
...
+  - ``--ft-attrsvc-port <PORT>`` (alias: ``--ft_attrsvc_port``), default ``50050`` when attribution is enabled
```
namitdhameja: --ft-attribution-port

docs/source/fault_tolerance/usage_guide.rst
```diff
@@ -112,30 +112,59 @@ Validation behavior:
...
+  - ``--ft-attrsvc-applog-dir <DIR>`` (alias: ``--ft_attrsvc_applog_dir``), default ``dirname(realpath(--ft-per-cycle-applog-prefix))``
```
namitdhameja: isn't this internally handled, do we need this to be explicitly provided? Or is this for the case when the cycle log is not activated.

pyproject.toml
```diff
@@ -52,16 +54,37 @@ langchain-openai = { version = ">=0.3.0", optional = true }
 mcp = { version = ">=1.15.0", optional = true }
 setproctitle = { version = ">=1.3.0", optional = true }
 logsage = { version = ">=0.1.7", optional = true }
+fastapi = { version = ">=0.100.0", optional = true }
```
namitdhameja, line 56: doesn't this take away from having attribution as an optional module? ie now pretty much all of the libs that are only required by attribution are also required in mainline.

pyproject.toml
```diff
@@ -52,16 +54,37 @@ langchain-openai = { version = ">=0.3.0", optional = true }
...
+slack-bolt = { version = ">=1.23.0", optional = true }
+slack-sdk = { version = ">=3.35.0", optional = true }
 grpcio = "^1.76.0"
 grpcio-tools = "^1.76.0"
 protobuf = ">=4.22.0"
 
 [tool.poetry.scripts]
 ft_launcher = "nvidia_resiliency_ext.fault_tolerance.launcher:main"
 nvrx-mcp-analysis = "nvidia_resiliency_ext.attribution.mcp_integration.server_launcher:main"
+nvrx-attrsvc = "nvidia_resiliency_ext.services_entrypoints:attrsvc_main"
```
namitdhameja: you only want to include these, if the user requires attribution so this should not be in the default

pyproject.toml
```diff
@@ -52,16 +54,37 @@ langchain-openai = { version = ">=0.3.0", optional = true }
...
+nvrx-smonsvc = "nvidia_resiliency_ext.services_entrypoints:smonsvc_main"
```
namitdhameja: smon is not required at all for ft

docs/source/fault_tolerance/usage_guide.rst
```diff
@@ -112,30 +112,59 @@ Validation behavior:
 Attribution service integration
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Enable artifact analysis (e.g., logs) during rendezvous health checks by pointing to a running attribution service.
-The feature is enabled by specifying both host and port.
+Per-cycle application logs do not enable attribution by themselves. To enable attribution, set
+``--ft-attrsvc-host``. A local host makes ``ft_launcher`` run ``nvrx_attrsvc`` on the TCPStore
```
namitdhameja: host implies IP. Perhaps leave the window open for UDS as well. We might only support IP for now or ever, but no reason to not have support for others. uri is an option as well '--ft-attribution-uri'?

namitdhameja (changes requested):

pyproject.toml
hexinw-nvidia: Attribution is optional. All these logsage and mcp dependencies are not involved if the user runs only "pip install nvidia_resiliency_ext*.whl".

pyproject.toml
hexinw-nvidia: For one wheel, nvrx-attrsvc will always be installed. But the plain installation (i.e. pip install nvidia_resiliency_ext*.whl) does not pull in the attribution dependencies. 

There is no standard one-wheel solution where pip install wheel skips the nvrx-attrsvc files/script but pip install 'wheel[attribution]' installs them.

docs/source/fault_tolerance/usage_guide.rst
hexinw-nvidia: I replaced "--ft-attrsvc-host" and "--ft-attrsvc-port" to "--ft-attribution-endpoint".

docs/source/fault_tolerance/usage_guide.rst
hexinw-nvidia: With "--ft-attribution-endpoint", "--ft-attribution-port" is not needed anymore.

docs/source/fault_tolerance/usage_guide.rst
hexinw-nvidia: Removed.

pyproject.toml
hexinw-nvidia: nvrx-attrsvc is packaged and installed to under nvidia_resiliency_ext/attribution/nvrx-attrsvc, so it won't pollute the top level Python site_packages.

docs/source/fault_tolerance/usage_guide.rst
hexinw-nvidia: Used "--ft-attribution-endpoint"

namitdhameja (approved):

PR #319, opened 2026-05-02
Expose attrsvc restart recommendations

namitdhameja:
Adds a normalized attrsvc `recommendation` contract and shared parser.
Clients now use `recommendation.action` (`STOP`, `RESTART`, `CONTINUE`, `UNKNOWN`, `TIMEOUT`) instead of interpreting raw `result.state`. Updates ft_launcher client parsing, smonsvc output handling, docs, and focused tests.

### DIFF
--- a/src/nvidia_resiliency_ext/attribution/orchestration/client_response.py
+++ b/src/nvidia_resiliency_ext/attribution/orchestration/client_response.py
@@ -1,0 +1,184 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Client-side helpers for attrsvc response payloads."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Any
+
+from .config import (
+    RESP_LOG_FILE,
+    RESP_MODE,
+    RESP_MODULE,
+    RESP_RESULT,
+    RESP_RESULT_ID,
+    RESP_SCHED_RESTARTS,
+    RESP_STATUS,
+    RESP_WL_RESTART,
+)
+from .llm_output import attribution_recommendation
+from .types import (
+    RECOMMENDATION_CONTINUE,
+    RECOMMENDATION_RESTART,
+    RECOMMENDATION_STOP,
+    RECOMMENDATION_TIMEOUT,
+    RECOMMENDATION_UNKNOWN,
+    AttributionRecommendation,
+)
+
+_VALID_RECOMMENDATION_ACTIONS = {
+    RECOMMENDATION_STOP,
+    RECOMMENDATION_RESTART,
+    RECOMMENDATION_CONTINUE,
+    RECOMMENDATION_UNKNOWN,
+    RECOMMENDATION_TIMEOUT,
+}
+_SPLITLOG_MODE = "splitlog"
+
+
+@dataclass(frozen=True)
+class AttrSvcResult:
+    """Normalized view of an attrsvc ``GET /logs`` response."""
+
+    result: Any
+    status: str = "completed"
+    log_path: str | None = None
+    recommendation: AttributionRecommendation = field(default_factory=AttributionRecommendation)
+    should_stop: bool = False
+    mode: str = "single"
+    analyzed_log_file: str = ""
+    wl_restart: Any = None
+    sched_restarts: Any = None
+
+    @property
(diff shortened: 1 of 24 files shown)


hexinw-nvidia (commented):
LGTM. 
I'd let Haim and Chengsong take another pass.

namitdhameja:
Addressed the remaining `_extract_source` P2 by ignoring blank/whitespace-only `module` values and falling through to the nested result module when present. Added a unit test for that case.

For the `health_check.py` sync/async GET race finding: I think this is a false positive for the current attrsvc implementation. The ft_launcher production path uses `_submit_log(...)` plus the fire-and-forget `AttributionService.__call__()` path; `get_last_result()` has no production call sites today. Even if a future caller used both for the same log, attrsvc GET analysis goes through `RequestCoalescer.get_or_compute()`, which implements single-flight behavior per log key. If a compute is already in flight, the second GET waits on the existing future rather than starting another compute; after completion, the result is cached.

So the worst case is duplicate client GET requests/log lines, not duplicate compute or incorrect stop/restart behavior. Leaving that path unchanged.

src/nvidia_resiliency_ext/attribution/orchestration/llm_output.py
```diff
@@ -20,51 +20,152 @@
...
+    RECOMMENDATION_STOP,
+    RECOMMENDATION_TIMEOUT,
+    RECOMMENDATION_UNKNOWN,
+    AttributionRecommendation,
+)
 
 logger = logging.getLogger(__name__)
 
 
 def attribution_no_restart(attr_result: Optional[Dict[str, Any]]) -> bool:
```
helisha91: @namitdhameja 
Are you using this function?

src/nvidia_resiliency_ext/attribution/orchestration/client_response.py
```diff
@@ -0,0 +1,183 @@
...
+                f"  Mode: {_SPLITLOG_MODE} (wl_restart {self.wl_restart}/{self.sched_restarts})"
+            )
+            if self.log_path:
+                lines.append(f"  Slurm output: {self.log_path}")
+            if self.analyzed_log_file:
+                lines.append(f"  Analyzed log: {self.analyzed_log_file}")
+        elif self.log_path:
+            lines.append(f"  Log: {self.log_path}")
+
+        if self.module:
```
helisha91, line 106: self.module is logsage or fr?

src/nvidia_resiliency_ext/services/attrsvc/routes.py
```diff
@@ -1,20 +1,22 @@
...
-PARAM_LOG_PATH = "log_path"
-PARAM_USER = "user"
-PARAM_JOB_ID = "job_id"
-
-# GET /logs optional query parameters (splitlog mode)
-PARAM_FILE = "file"
-PARAM_WL_RESTART = "wl_restart"
+__all__ = [
+    "PARAM_FILE",
+    "PARAM_JOB_ID",
```
helisha91, line 17: In next PR, we probably want to add cycle_id

src/nvidia_resiliency_ext/attribution/orchestration/llm_output.py
```diff
@@ -20,51 +20,152 @@
...
 from typing import Any, Dict, Optional
 
+from .config import RESP_ERROR, RESP_MODULE, RESP_RESULT, RESP_STATE, STATE_TIMEOUT
 from .log_path_metadata import JobMetadata
+from .types import (
+    RECOMMENDATION_CONTINUE,
+    RECOMMENDATION_RESTART,
+    RECOMMENDATION_STOP,
+    RECOMMENDATION_TIMEOUT,
+    RECOMMENDATION_UNKNOWN,
```
helisha91, line 30: @namitdhameja 
What is unknown case?

namitdhameja: Its attribution completed without a usable action, used for FR

src/nvidia_resiliency_ext/attribution/orchestration/client_response.py
namitdhameja: yes logsage or fr

src/nvidia_resiliency_ext/services/attrsvc/routes.py
namitdhameja: the fully specified log path has the cycle id already

hexinw-nvidia (approved):

PR #321, opened 2026-05-03
Add external InJob control plane support

hexinw-nvidia:
Restructure InJob fault-tolerance control so the same rendezvous protocol can run in two deployment modes:

- colocated control: control-plane duties run inside the compute-side ft_launcher on the rendezvous/TCPStore host
- external control: control-plane duties run in a dedicated nvrx-control process outside the compute launcher/job

The control plane owns the long-lived coordination services for an InJob run:
- TCPStore hosting
- rendezvous coordination
- gRPC log aggregation
- Managed attribution service interaction
- cycle-info reporting including rank assignment

Add the nvrx-control entrypoint for external control. The external control process closes compute-only rendezvous rounds without joining as a training participant, so compute rank assignment, WORLD_SIZE, standby assignment, and worker launch remain compute-only.

Make external control independent of SLURM environment. Cycle-info file naming uses an explicit --ft-cycle-info-job-id with attempt_index=0, while colocated control continues to derive its cycle-info namespace from the local SLURM rank0 environment. This allows nvrx-control to run on a CPU node that does not have SLURM_JOB_ID, SLURM_PROCID, or SLURM_RESTART_CNT set.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,6 +66,7 @@ protobuf = ">=4.22.0"
 
 [tool.poetry.scripts]
 ft_launcher = "nvidia_resiliency_ext.fault_tolerance.launcher:main"
+nvrx-control = "nvidia_resiliency_ext.fault_tolerance.control_plane:main"
 nvrx-mcp-analysis = "nvidia_resiliency_ext.attribution.mcp_integration.server_launcher:main"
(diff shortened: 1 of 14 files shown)


hexinw-nvidia:
@greptile

namitdhameja (approved):

PR #325, opened 2026-05-06
Add previous-active last-call gate for hot spares

hexinw-nvidia:
Delay closing barrier rendezvous restart rounds when hot spares arrive before previous active nodes. The rendezvous host now remembers the last closed round's active infrastructure ranks locally and, when max_nodes is greater than min_nodes, starts a bounded last_call window after the segment constraint is satisfied. If previous active nodes rejoin before the window expires, they participate in the final topology-based assignment; otherwise the host closes with the available snapshot.

This keeps rank assignment topology-driven rather than filling old rank holes explicitly. The tradeoff is intentional: a slow previous active node can still be replaced after last_call_timeout, but early hot spares no longer win simply because they reached rendezvous first. The design avoids a lease/peer-health protocol and avoids persisting previous-active state in TCPStore because the barrier path assumes a single rendezvous host.

Default last_call_timeout to 15 seconds and wire it through launcher and control-plane rendezvous config.


hexinw-nvidia:
@greptile

namitdhameja (approved):

PR #329, opened 2026-05-11
Fix progress tracker YAML values overwritten by CLI defaults

chenming1986:
## Summary

Fix progress tracker config precedence so YAML values are not overwritten when the corresponding CLI args are omitted.

`FaultToleranceConfig.from_args()` loads YAML config first, then applies parsed CLI args when their value is not `None`. The progress tracker CLI args had argparse numeric defaults, so omitted CLI args still overwrote YAML values with those defaults.

This changes the progress tracker CLI defaults to `None`, matching the convention used by other YAML-configurable args. Dataclass defaults still apply when neither YAML nor CLI provides a value, and explicit CLI values continue to override YAML.

## Testing

- Added unit coverage for preserving YAML progress tracker values when CLI args are absent
- `python -m pytest tests/fault_tolerance/unit/test_config.py`: 16 passed

### DIFF
--- a/src/nvidia_resiliency_ext/fault_tolerance/launcher.py
+++ b/src/nvidia_resiliency_ext/fault_tolerance/launcher.py
@@ -2545,7 +2545,7 @@ def get_args_parser() -> ArgumentParser:
         "--ft-max-no-progress-cycles",
         "--ft-max_no_progress_cycles",
         type=int,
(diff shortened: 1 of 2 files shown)


hexinw-nvidia (approved):
Thanks for the fix.

PR #333, opened 2026-05-12
[codex] Enable ft launcher attrsvc observability

namitdhameja:
## Summary

- Add ft_launcher plumbing for launcher-managed attrsvc dataflow posting via explicit `--ft-attribution-dataflow-url` / `attribution_dataflow_url` configuration.
- Set the managed child service `NVRX_ATTRSVC_DATAFLOW_URL` only from that explicit launcher config, and strip any parent env value so launcher does not configure dataflow from env.
- Keep the dev feature on the new canonical env/flag names only, with no old-name compatibility path.
- Pass launcher environment-derived `user` and `job_id` metadata on attrsvc `POST /logs` submissions, using shared job metadata helpers.
- Document the dataflow URL configuration and launcher job metadata behavior using public-safe example URLs.

## Impact

Managed ft_launcher attribution can now post results to a configured dataflow endpoint and include job/user metadata needed by attrsvc for splitlog tracking and dataflow records.

## Validation

- `compileall` on touched Python files
- `git diff --check`
- Confirmed old env/flag/config spellings are absent under `src`, `services`, `tests`, and `docs`
- Confirmed internal dataflow hostname/index strings are absent under the repo
- `black --check .`, `isort --check-only .`, `ruff check .`
- `tests/attribution/unit/test_post_backend.py`, `tests/attribution/unit/test_postprocessing_pipeline.py`, `tests/shared_utils/test_health_check.py::TestAttributionService`, `tests/shared_utils/test_job_metadata.py`
- `tests/fault_tolerance/unit/test_attribution_manager.py` under a local import harness because this local Python environment lacks the full torch runtime
- Focused launcher CLI/config verification for `--ft-attribution-dataflow-url`

## Notes

- `tests/fault_tolerance/unit/test_cycle_info_writer.py` could not be run in this local checkout because generated protobuf modules are absent, but the touched cycle-info code path now delegates to the separately tested shared job-id helper.

### DIFF
--- a/docs/source/fault_tolerance/usage_guide.rst
+++ b/docs/source/fault_tolerance/usage_guide.rst
@@ -230,6 +230,7 @@ service dependencies.
   - ``--ft-attribution-llm-base-url <URL>`` (alias: ``--ft_attribution_llm_base_url``)
   - ``--ft-attribution-llm-model <MODEL>`` (alias: ``--ft_attribution_llm_model``)
   - ``--ft-attribution-startup-timeout <SECONDS>`` (alias: ``--ft_attribution_startup_timeout``), default ``20``
+  - ``--ft-attribution-export-url <URL>`` (alias: ``--ft_attribution_export_url``)
 
   The managed attribution app-log directory is derived from
   ``dirname(realpath(--ft-per-cycle-applog-prefix))``. Its stdout/stderr log is derived
@@ -239,6 +240,14 @@ service dependencies.
   The managed attribution API key must come from ``--ft-attribution-llm-api-key-file`` or inherited
   ``LLM_API_KEY_FILE``. If neither points to a readable file, the TCPStore-host launcher fails
   before starting the attribution service.
+
+  To export managed attribution results, pass ``--ft-attribution-export-url`` or
+  set ``attribution_export_url`` in the fault tolerance YAML config.
+
+  ``ft_launcher`` sends job metadata with each attribution submission: ``user`` is read from
+  ``SLURM_JOB_USER`` or ``USER``, and ``job_id`` is read from ``SLURM_ARRAY_JOB_ID`` or
+  ``SLURM_JOB_ID``. If no corresponding environment variable is set, that field is omitted from
+  the submission payload.
 
   Example:
 
@@ -250,6 +259,7 @@ service dependencies.
        --ft-attribution-llm-api-key-file /secure/llm_api_key \
        --ft-attribution-llm-base-url https://integrate.api.nvidia.com/v1 \
        --ft-attribution-llm-model nvidia/nemotron-3-super-120b-a12b \
+       --ft-attribution-export-url https://dataflow.example.test/dataflow2/example-index/posting \
        train.py
 
   To use an externally managed attribution service instead, specify an explicit endpoint:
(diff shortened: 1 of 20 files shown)


hexinw-nvidia (commented):
.

hexinw-nvidia (commented):
Review notes (not blocking). Numbered for easy reply.

**1. Naming — `dataflow` leaks internal package name into public CLI/YAML/env.**
`nvdataflow` is the implementation; the user-facing role is an ingest/results sink. Suggest `--ft-attribution-ingest-url` / `attribution_ingest_url` / `NVRX_ATTRSVC_INGEST_URL`. Since the env var is *already* being renamed here, this is the cheap moment to do it; keep `nvdataflow` only inside `post_backend.py`.

**2. Breaking env rename with no deprecation shim.**
`NVRX_ATTRSVC_DATAFLOW_HTTP_URL` → `NVRX_ATTRSVC_DATAFLOW_URL` is a hard break for any existing deployment (systemd units, k8s configmaps outside this repo). If pre-1.0 / no external consumers, fine; otherwise accept both for one release and log a deprecation warning on the old name.

**3. CLI flag silently dropped for external endpoints.**
`AttributionConfig.from_args` only calls `_resolve_dataflow_url` inside the `is_managed_attribution_endpoint(endpoint)` branch. If a user passes `--ft-attribution-dataflow-url` together with an external `--ft-attribution-endpoint`, the flag is discarded without warning. `test_external_attribution_ignores_launcher_dataflow_env` confirms the env case but not the explicit CLI case. Either raise on the combo or log a warning — silent drop will confuse operators.

**4. Empty-string vs omitted `user` / `job_id`.**
`job_user_from_env()` / `job_id_from_env()` always return `str` (never `None`), and `log_submit_payload` only omits keys whose value is `None`. So when SLURM env is unset the POST body carries `"user": ""`, `"job_id": ""` — see `test_http_endpoint_posts_empty_job_metadata_when_env_unset`. If attrsvc treats `""` identically to absence, fine; otherwise return `None` from the helpers (or pass `None` at the call site) so the field is omitted. Worth confirming on the server side.

**5. Nit — redundant trailing `or ""` in helpers.**
`shared_utils/job_metadata.py`:
- `return os.environ.get("SLURM_JOB_USER") or os.environ.get("USER", "") or ""` — the trailing `or ""` is dead; `os.environ.get(key, "")` already returns `""`.
- Same for `job_id_from_env`.

**6. Doc nit — `docs/source/fault_tolerance/usage_guide.rst`.**
The new paragraph documents that `user`/`job_id` are read from SLURM env, but doesn't say what happens when those vars are unset (empty strings are sent, see #4). Worth a one-liner.

**Strengths**
- Parent-env scrub of `NVRX_ATTRSVC_DATAFLOW_URL` in `_child_env` is exactly right.
- CLI > YAML > env precedence is symmetric with sibling fields and well-tested.
- Validator rejects non-http(s), missing netloc, query, fragment — appropriate given the queue param is appended internally.
- `job_metadata.py` extraction DRYs three identical SLURM env-id fallbacks.
- Test coverage on validator, override paths, child-env presence/absence, payload shape, and helper fallbacks is solid.

hexinw-nvidia (approved):

PR #338, opened 2026-05-15
[codex] Add ft launcher attribution integration

namitdhameja:
## Summary
- builds on merged #339 service/smonsvc terminal attribution support
- add launcher-managed attribution restart decisions and host-managed attribution service wiring
- move the final STOP decision to the rendezvous host before closing the next round, so launcher only signals terminal attribution early
- keep attribution shutdown state simple: the rendezvous host exits gracefully when it decides STOP; other ranks just observe rendezvous closure
- reframe attribution timeout as a launcher-side decision budget; terminal POST starts the budget, each GET is a short local request, and the rendezvous host reruns its normal close loop until a decision arrives or the budget expires
- record attribution GET wait profiling events, including completed-event duration_seconds metadata

## Validation
- compileall on changed launcher/rendezvous/shared-utils files and tests
- git diff --check
- isort --check-only on changed Python files
- black --check on changed Python files

Note: focused pytest could not run in the local desktop runtime because required test dependencies are not installed there.

### DIFF
--- a/src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
+++ b/src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
@@ -630,6 +630,8 @@ class _RendezvousBarrierState:
         self.prefix = f"ft_rendezvous_barrier:{run_id}"
         # Permanent shutdown: once set, no further rendezvous rounds (graceful exit).
         self.shutdown_key = f"{self.prefix}:shutdown"
+        self._attribution_service: Optional[AttributionService] = None
+        self._attribution_settled_for_round: int = -1
         # Job-level unhealthy counter; persists across all rounds.
         self.unhealthy_count_key = f"{self.prefix}:unhealthy_count"
         # Latest closed round's active membership. This is host-local by design:
@@ -1250,7 +1252,7 @@ class _RendezvousBarrierState:
             node_desc: Node descriptor for logging
 
         Raises:
-            RendezvousClosedError: If rendezvous was permanently shut down
+            RendezvousClosedError: If rendezvous was permanently shut down.
             RendezvousTimeoutError: If rendezvous has timed out
         """
         # Check for permanent shutdown
@@ -1349,6 +1351,35 @@ class _RendezvousBarrierState:
             wait_count += 1
             time.sleep(1.0)  # Poll every 1 second
 
+    def _attribution_gate(self, node_desc: _NodeDesc) -> bool:
+        """Return whether attribution permits closing the current round.
+
+        Raises ``RendezvousGracefulExitError`` if attribution recommends stopping.
+        Returns ``False`` while attribution is still pending.
+        """
+        if self._attribution_service is None or self._round == 0:
+            return True
+
+        if self._attribution_settled_for_round == self._round:
+            return True
+
+        result = self._attribution_service.get_last_result(node_id=node_desc)
+
+        if result is None:
+            return False
+
+        if result is True:
+            msg = (
+                f"The node '{node_desc}' is closing rendezvous because attribution "
+                f"recommended stopping the job."
+            )
+            log.error(msg)
+            self.set_shutdown()
+            raise RendezvousGracefulExitError(msg)
+
+        self._attribution_settled_for_round = self._round
+        return True
+
     def _host_close_round(
         self,
         node_desc: _NodeDesc,
@@ -1406,6 +1437,10 @@ class _RendezvousBarrierState:
 
             # Only attempt snapshot + constraint check when enough nodes have joined.
             if current_joined < min_nodes:
+                time.sleep(0.1)
+                continue
+
+            if not self._attribution_gate(node_desc):
                 time.sleep(0.1)
                 continue
 
@@ -2085,6 +2120,7 @@ class FtRendezvousBarrierHandler(RendezvousHandler):
         link_state_path_template: Optional[str] = None,
         storage_healthcheck_paths: Optional[list] = None,
         attribution_endpoint: Optional[str] = None,
+        attribution_decision_timeout: Optional[float] = None,
         cycle_info_dir: Optional[str] = None,
         cycle_log_prefix: Optional[str] = None,
     ):
@@ -2121,6 +2157,8 @@ class FtRendezvousBarrierHandler(RendezvousHandler):
                 List of storage paths to check for health.
             attribution_endpoint:
                 Endpoint of the attribution service.
+            attribution_decision_timeout:
+                Launcher-side attribution decision budget in seconds.
         """
         # We associate each handler instance with a unique node descriptor.
         node = cls._node_desc_generator.generate(local_addr)
@@ -2148,6 +2186,7 @@ class FtRendezvousBarrierHandler(RendezvousHandler):
             link_state_path_template=link_state_path_template,
             storage_healthcheck_paths=storage_healthcheck_paths,
             attribution_endpoint=attribution_endpoint,
+            attribution_decision_timeout=attribution_decision_timeout,
             cycle_info_dir=cycle_info_dir,
             cycle_log_prefix=cycle_log_prefix,
         )
@@ -2164,6 +2203,7 @@ class FtRendezvousBarrierHandler(RendezvousHandler):
         link_state_path_template: Optional[str] = None,
         storage_healthcheck_paths: Optional[list] = None,
         attribution_endpoint: Optional[str] = None,
+        attribution_decision_timeout: Optional[float] = None,
         cycle_info_dir: Optional[str] = None,
         cycle_log_prefix: Optional[str] = None,
     ) -> None:
@@ -2239,9 +2279,13 @@ class FtRendezvousBarrierHandler(RendezvousHandler):
 
         # Attribution service client (optional, only on master node)
         if is_store_host and attribution_endpoint:
-            self._attribution_service = AttributionService(endpoint=attribution_endpoint)
+            self._attribution_service = AttributionService(
+                endpoint=attribution_endpoint,
+                decision_timeout=attribution_decision_timeout,
+            )
         else:
             self._attribution_service = None
+        self._barrier_state._attribution_service = self._attribution_service
 
     @property
     def _rendezvous_round(self) -> int:
@@ -2387,11 +2431,6 @@ class FtRendezvousBarrierHandler(RendezvousHandler):
                 "Storage path health check",
                 f"Node {self._this_node} has invalid or unreadable paths.",
             )
-
-        # Perform optional log analysis (non-fatal)
-        # Note: _submit_log() was already called from launcher before workers started
-        if self._attribution_service is not None:
-            self._attribution_service()
 
         # Perform Node health check (external service if available)
         _nodehealth_checker = get_node_health_check()
@@ -2784,6 +2823,9 @@ def create_handler(
         storage_healthcheck_paths = params.config.get('storage_healthcheck_paths', None)
         link_state_path_template = params.config.get('link_state_path_template', None)
         attribution_endpoint = params.config.get('attribution_endpoint', None)
+        attribution_decision_timeout = params.config.get('attribution_decision_timeout', None)
+        if attribution_decision_timeout is not None:
+            attribution_decision_timeout = float(attribution_decision_timeout)
         cycle_info_dir = params.config.get('cycle_info_dir', None)
         cycle_log_prefix = params.config.get('cycle_log_prefix', None)
 
@@ -2804,6 +2846,7 @@ def create_handler(
             link_state_path_template=link_state_path_template,
             storage_healthcheck_paths=storage_healthcheck_paths,
             attribution_endpoint=attribution_endpoint,
+            attribution_decision_timeout=attribution_decision_timeout,
             cycle_info_dir=cycle_info_dir,
             cycle_log_prefix=cycle_log_prefix,
         )
--- a/src/nvidia_resiliency_ext/fault_tolerance/launcher.py
+++ b/src/nvidia_resiliency_ext/fault_tolerance/launcher.py
@@ -475,7 +475,6 @@ class LocalElasticAgent(SimpleElasticAgent):
         """
         return self._rdzv_handler.round()
 
-
     def _current_cycle_info_path(self) -> Optional[str]:
         if not self._ft_cfg.cycle_info_dir:
             return None
@@ -536,7 +535,7 @@ class LocalElasticAgent(SimpleElasticAgent):
         log_msg: str,
         open_rendezvous: bool = False,
     ) -> bool:
-        """Handle restart decision logic based on progress tracking and remaining restarts.
+        """Handle restart decision logic based on attribution, progress, and restart budget.
 
         Args:
             role: The role name for logging
@@ -548,6 +547,7 @@ class LocalElasticAgent(SimpleElasticAgent):
             True if restart was initiated (caller should continue monitoring loop)
             False if no restart (caller should stop workers and return failure)
         """
+        self._request_terminal_attribution()
         self._progress_tracker.analyze_previous_cycle()
         should_terminate_early = self._progress_tracker.should_terminate_early()
 
@@ -867,6 +867,15 @@ class LocalElasticAgent(SimpleElasticAgent):
         #       The 'name' field of the Event is NOT used in the TorchelasticStatusLogEntry.
         event = events.Event(name=name, source=events.EventSource.AGENT, metadata=metadata)
         events.record(event)
+
+    def _request_terminal_attribution(self) -> None:
+        """Ask attrsvc to start final analysis for the last submitted cycle log."""
+        if not self._is_store_host:
+            return
+        attribution_service = getattr(self._rdzv_handler, "_attribution_service", None)
+        if attribution_service is None:
+            return
+        attribution_service.request_terminal_analysis()
 
     # pyre-fixme[56]: Pyre was not able to infer the type of the decorator
     #  `torch.distributed.elastic.metrics.prof`.
@@ -2790,7 +2799,6 @@ def _validate_attribution_requires_per_cycle_applog(
             "Attribution service needs per-cycle application logs as analysis input."
         )
 
-
 def config_from_args(args, launcher_pipe_read_fd=None, launcher_log_file=None) -> Tuple[LaunchConfig, Union[Callable, str], List[str]]:
     # If ``args`` not passed, defaults to ``sys.argv[:1]``
     _validate_args(args)
@@ -2871,6 +2879,7 @@ def config_from_args(args, launcher_pipe_read_fd=None, launcher_log_file=None) -
     # Pass enable_nic_healthcheck and link_state_path_template from fault tolerance config to rendezvous config
     rdzv_configs['enable_nic_healthcheck'] = fault_tol_cfg.enable_nic_healthcheck
     rdzv_configs['link_state_path_template'] = fault_tol_cfg.link_state_path_template
+
     # Pass distributed storage health check configuration
     cli_dist_storage = getattr(args, 'ft_enable_dist_storage_healthcheck', None)
     if cli_dist_storage is not None:
@@ -2972,6 +2981,11 @@ def config_from_args(args, launcher_pipe_read_fd=None, launcher_log_file=None) -
         attribution_endpoint = attribution_manager.start_if_needed()
         if attribution_endpoint is not None:
             rdzv_configs['attribution_endpoint'] = attribution_endpoint.endpoint
+            decision_timeout = getattr(attribution_endpoint, "decision_timeout", None)
+            if decision_timeout is not None:
+                rdzv_configs['attribution_decision_timeout'] = str(
+                    decision_timeout
+                )
 
         if getattr(args, 'ft_enable_log_server', False):
             assert log_funnel_ports is not None
--- a/src/nvidia_resiliency_ext/shared_utils/health_check.py
+++ b/src/nvidia_resiliency_ext/shared_utils/health_check.py
@@ -23,10 +23,11 @@ import os
 import subprocess  # nosec B404
 import sys
 import threading
+import time
 import traceback
 from collections import defaultdict, deque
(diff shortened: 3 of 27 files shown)


src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -630,6 +632,10 @@ def __init__(
         self.prefix = f"ft_rendezvous_barrier:{run_id}"
         # Permanent shutdown: once set, no further rendezvous rounds (graceful exit).
         self.shutdown_key = f"{self.prefix}:shutdown"
+        # Optional shutdown reason lets non-host nodes map attribution STOP shutdowns
+        # to RendezvousGracefulExitError instead of a generic closed-rendezvous error.
+        self.shutdown_reason_key = f"{self.prefix}:shutdown_reason"
```
hexinw-nvidia: "shutdown_reason" key is not needed.

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -1250,13 +1256,16 @@ def _check_timeout_and_closure(self, node_desc: _NodeDesc) -> None:
...
-            RendezvousClosedError: If rendezvous was permanently shut down
+            RendezvousGracefulExitError: If attribution requested a job-level stop.
+            RendezvousClosedError: If rendezvous was permanently shut down for other reasons.
             RendezvousTimeoutError: If rendezvous has timed out
         """
         # Check for permanent shutdown
         if self.is_shutdown():
             msg = f"The node '{node_desc}' detected that rendezvous was permanently closed"
             log.info(msg)
+            if self._shutdown_reason() == _SHUTDOWN_REASON_ATTRIBUTION_STOP:
```
hexinw-nvidia: Not needed as discussed.

src/nvidia_resiliency_ext/fault_tolerance/launcher.py
```diff
@@ -2781,15 +2790,15 @@ def _resolve_attribution_endpoint(args: Any, fault_tol_cfg: FaultToleranceConfig
 def _validate_attribution_requires_per_cycle_applog(
     args: Any, fault_tol_cfg: FaultToleranceConfig
 ) -> None:
-    if _resolve_attribution_endpoint(args, fault_tol_cfg) and not getattr(
-        args, "ft_per_cycle_applog_prefix", None
-    ):
+    if getattr(args, "ft_per_cycle_applog_prefix", None):
+        return
+
+    if _resolve_attribution_endpoint(args, fault_tol_cfg):
```
hexinw-nvidia: The original check is correct. Revert this change.

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -1530,6 +1556,7 @@ def _host_close_round(
                 f"participants{replacement_group_info}{unhealthy_replacement_group_info}, "
                 f"min={min_nodes} (fetch {fetch_elapsed*1000:.1f}ms, check {check_elapsed*1000:.1f}ms)"
             )
+            self._check_attribution_before_round_close(node_desc)
```
hexinw-nvidia: This should be moved to before the constraint check, as discussed.

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -37,12 +37,19 @@
...
+    ANALYSIS_INTENT_TERMINAL,
+)
 from nvidia_resiliency_ext.shared_utils.job_metadata import job_id_from_env, job_user_from_env
 from nvidia_resiliency_ext.shared_utils.log_manager import LogConfig
 
 # Get the nvrx logger
 logger = logging.getLogger(LogConfig.name)
 
+_ATTRIBUTION_POST_TIMEOUT_SECONDS = 2.0
+_ATTRIBUTION_GET_TIMEOUT_SECONDS = 5.0
```
hexinw-nvidia: Change to 500 msec.

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -1377,6 +1404,9 @@ def _host_close_round(
         mismatch_first_seen: Optional[float] = None
         close_gate = self._make_previous_active_last_call_gate(min_nodes, max_nodes)
         last_call_logged_deadline: Optional[float] = None
+        # Round 0 has no prior workload logs to attribute. Restart rounds wait for
+        # attribution once, then re-enter the loop so close checks use fresh state.
+        attribution_close_allowed = self._attribution_service is None or self._round == 0
```
hexinw-nvidia: Move the attribution gate behind a single self-contained helper, with no local flag in the loop:

```suggestion
      def _attribution_gate(self, node_desc) -> bool:                                    
      """True if close is permitted by attribution. Raises if STOP. False if still pending"""
      if self._attribution_service is None or self._round == 0:                      
          return True
      if self._attribution_settled_for_round == self._round:
          return True                                                                
      result = self._attribution_service.get_last_result(node_id=node_desc)
      if result is None:                                                             
          return False
      if result is True:                                                             
          log.error(f"node '{node_desc}' closing rendezvous: attribution recommended stop")
          self.set_shutdown()                                                        
          raise RendezvousGracefulExitError(...)
      self._attribution_settled_for_round = self._round
      return True                    
```

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -1409,6 +1439,10 @@ def _host_close_round(
                 time.sleep(0.1)
                 continue
 
+            if not attribution_close_allowed:
+                attribution_close_allowed = self._check_attribution_before_round_close(node_desc)
+                continue
+
```
hexinw-nvidia: This becomes:

      if not self._attribution_gate(node_desc):
          time.sleep(0.1)   # explicit pacing instead of relying on HTTP timeout     
          continue

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -630,6 +630,7 @@ def __init__(
         self.prefix = f"ft_rendezvous_barrier:{run_id}"
         # Permanent shutdown: once set, no further rendezvous rounds (graceful exit).
         self.shutdown_key = f"{self.prefix}:shutdown"
+        self._attribution_service: Optional[AttributionService] = None
```
hexinw-nvidia, line 633: You can initialize the "_attribution_settled_for_round" here.

```suggestion
        self._attribution_service: Optional[AttributionService] = None
        self._attribution_settled_for_round: int = -1
```

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -1349,6 +1350,32 @@ def _wait_for_rendezvous_open(
             wait_count += 1
             time.sleep(1.0)  # Poll every 1 second
 
+    def _check_attribution_before_round_close(self, node_desc: _NodeDesc) -> bool:
```
hexinw-nvidia: Replace with the _attribution_gate().

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -1358,87 +1366,224 @@ class AttributionService:
...
+        if self._terminal_deadline is None:
+            self._terminal_deadline = time.monotonic() + self.decision_timeout
+        return self._terminal_deadline
+
+    def _remaining_decision_timeout(self) -> float:
+        """Return the remaining launcher-side budget for the current attribution decision."""
+        remaining = self._ensure_decision_deadline() - time.monotonic()
+        if remaining <= 0:
+            logger.info(
+                "AttributionService attribution decision budget exhausted",
```
hexinw-nvidia: Drop this duplicate log. It fires every poll iteration once expired.

src/nvidia_resiliency_ext/shared_utils/profiling.py
```diff
@@ -20,7 +20,7 @@
 import time
 from datetime import datetime, timezone
 from enum import Enum
-from typing import Any, Optional
+from typing import Any, Mapping, Optional
```
hexinw-nvidia: Not needed. The "metadata" is not needed. The duration calculation will be done in post-processing tool.

src/nvidia_resiliency_ext/shared_utils/profiling.py
```diff
@@ -75,6 +77,7 @@ def record_event(
         event: ProfilingEvent,
         node_id: Optional[Any] = None,
         rank: Optional[int] = None,
+        metadata: Optional[Mapping[str, Any]] = None,
```
hexinw-nvidia: Not needed. Remove

src/nvidia_resiliency_ext/shared_utils/profiling.py
```diff
@@ -88,9 +91,14 @@ def record_event(
 
         # Format log message with cycle count and UTC time
         utc_time = self._timestamp_to_utc_datetime(timestamp)
+        metadata_msg = ""
+        if metadata:
+            metadata_msg = " " + " ".join(
+                f"{key}: {value}" for key, value in sorted(metadata.items())
+            )
```
hexinw-nvidia: Remove

src/nvidia_resiliency_ext/shared_utils/profiling.py
```diff
@@ -88,9 +91,14 @@ def record_event(
...
         self._logger.info(
             f"  - Cycle: {self._current_cycle} Event: {event.value} Node: {node_id_str} Rank: {rank} "
-            f"Time: {utc_time} UTC"
+            f"Time: {utc_time} UTC{metadata_msg}"
```
hexinw-nvidia: Undo

src/nvidia_resiliency_ext/shared_utils/profiling.py
```diff
@@ -115,18 +123,20 @@ def record_profiling_event(
     event: ProfilingEvent,
     node_id: Optional[Any] = None,
     rank: Optional[int] = None,
+    metadata: Optional[Mapping[str, Any]] = None,
```
hexinw-nvidia: Remove

src/nvidia_resiliency_ext/shared_utils/profiling.py
```diff
@@ -115,18 +123,20 @@ def record_profiling_event(
...
 ) -> str:
     """Convenience function to record a profiling event.
 
     Args:
         event: The profiling event to record
         node_id: Node identifier (can be any type, will be converted to string)
         rank: Rank identifier
+        metadata: Optional extra event fields to append to the profiling log line
```
hexinw-nvidia: remove

src/nvidia_resiliency_ext/shared_utils/profiling.py
```diff
@@ -115,18 +123,20 @@ def record_profiling_event(
...
 
     Returns:
         Event ID string
     """
-    return _get_global_profiler().record_event(event, node_id, rank)
+    return _get_global_profiler().record_event(event, node_id, rank, metadata)
```
hexinw-nvidia: remove

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -1358,87 +1366,224 @@ class AttributionService:
...
+    def _start_get_profiling(self, node_id: Optional[Any]) -> None:
+        if self._get_wait_start is not None:
+            return
+        self._get_wait_start = time.monotonic()
+        record_profiling_event(
+            ProfilingEvent.ATTRIBUTION_GET_STARTED,
+            node_id=node_id,
+        )
+
+    def _complete_get_profiling(
```
hexinw-nvidia: This method is not needed. Just need to report the ATTRIBUTION_GET_COMPLETED event.

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -1358,87 +1366,224 @@ class AttributionService:
...
-                target=self._get_results,
-                args=(log_path,),
-                daemon=True,
-            ).start()
-
-    def get_last_result(self) -> Optional[bool]:
-        """Synchronously fetch whether attribution recommends stopping the last log."""
+        self.request_terminal_analysis()
+
+    def get_last_result(
```
hexinw-nvidia, line 1819: Claude suggested rewriting this method.


```suggestion
  Suggested rewrite                                                                                                                                               

  After applying items 1–6:                                                          
 
  def get_last_result(self, node_id: Optional[Any] = None) -> Optional[bool]:
      """Fetch the attribution verdict for the last submitted workload log.          

      Returns:
          True   — attribution recommends STOP                                       
          False  — CONTINUE, or decision-budget expired (fail-open), or no submitted log
          None   — still pending within the decision budget                          
      """
      log_path = self._last_submitted                                                
      if not log_path:
          logger.debug("AttributionService GET skipped: no submitted log path")
          return False                                                               

      if not self._get_started_recorded:                                             
          record_profiling_event(ProfilingEvent.ATTRIBUTION_GET_STARTED, node_id=node_id)
          self._get_started_recorded = True                                          

      decision_remaining = self._remaining_decision_timeout()                        
      if decision_remaining <= 0:
          return self._fail_open_on_timeout(log_path, node_id)                       

      request_timeout = min(_ATTRIBUTION_REQUEST_TIMEOUT_SECONDS, decision_remaining)
      result = self._get_results(log_path, timeout=request_timeout)

      if result is None:
          if self._remaining_decision_timeout() <= 0:
              return self._fail_open_on_timeout(log_path, node_id)                   
          return None

      record_profiling_event(ProfilingEvent.ATTRIBUTION_GET_COMPLETED, node_id=node_id)
      return result

  def _fail_open_on_timeout(self, log_path: str, node_id: Optional[Any]) -> bool:    
      logger.warning(
          "AttributionService decision budget exhausted for %s (%.1fs); failing open",
          log_path, self.decision_timeout,
      )                                                                              
      record_profiling_event(ProfilingEvent.ATTRIBUTION_GET_COMPLETED, node_id=node_id)
      return False

  Plus _submit_log() resets self._get_started_recorded = False (replaces _get_wait_start = None).

  Net effect                                                                         
 
  - One parameter removed
  - One unreachable branch removed
  - Two helper methods removed
  - One duplicate log site eliminated                                                
  - Both timeout exits now log uniformly
  - Profile-event pair always balances (no orphan STARTED)
  - Method body shrinks ~40%
    
```

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -1349,6 +1351,36 @@ def _wait_for_rendezvous_open(
...
+            msg = (
+                f"The node '{node_desc}' is closing rendezvous because attribution "
+                f"recommended stopping the job."
+            )
+            log.error(msg)
+            self.set_shutdown()
+            raise RendezvousGracefulExitError(msg)
+
+        self._attribution_settled_for_round = self._round
+        return False
```
hexinw-nvidia: This should be "return True"? Why does the caller need to re-read the result?

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -1409,6 +1441,9 @@ def _host_close_round(
                 time.sleep(0.1)
                 continue
 
+            if not self._attribution_gate(node_desc):
+                continue
```
hexinw-nvidia, line 1445: Don't we need a time.sleep(0.1) or time.sleep(0.5) to prevent calling the get_last_result() back-to-back?

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -1358,87 +1366,209 @@ class AttributionService:
...
+        """
         log_path = self._last_submitted
         if not log_path:
             logger.debug("AttributionService GET skipped: no submitted log path")
+            return False
+
+        self._start_get_profiling(node_id)
+        decision_remaining = self._remaining_decision_timeout()
+        if decision_remaining <= 0:
+            logger.info(
```
hexinw-nvidia: Should we change it to logger.warning?

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -1358,87 +1366,209 @@ class AttributionService:
...
+                log_path,
+            )
+            record_profiling_event(
+                ProfilingEvent.ATTRIBUTION_GET_COMPLETED,
+                node_id=node_id,
+            )
+            return False
+
+        request_timeout = min(_ATTRIBUTION_REQUEST_TIMEOUT_SECONDS, decision_remaining)
+        if timeout is not None:
```
hexinw-nvidia: Is there any use case to pass in timeout in the get_last_result() call? If not, I think we can remove line 1428-1431

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
namitdhameja: That is intentional based on morning discussion. It forces one extra loop pass after attr is available so _host_close_round re-runs from the top.

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -1358,87 +1366,209 @@ class AttributionService:
...
+            return None
+
+        try:
+            result = self._get_results(log_path, timeout=request_timeout)
+        except Exception:
+            record_profiling_event(
+                ProfilingEvent.ATTRIBUTION_GET_COMPLETED,
+                node_id=node_id,
+            )
+            raise
```
hexinw-nvidia: try...except block is not needed. "_get_results" already catches the exception from the HTTP call. Correct?

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
namitdhameja: Whats the diff between GET(timeout=0.5) or GET(timeout=0.6) ? Latter is better as if the result is available at time interval 0.5-0.6, then it doesn't wait one more cycle and is satisfied immediately.

Though frankly, I don't get the benefit of breaking GET in multiple anyways. A GPU reclaim is 30 sec, a median DCA HC takes 25 sec. 55 sec is already done, what benefit does breaking GET for a 5 sec interval adds, whats the functionality that its breaking or has improved if we just wait for the budgeted time. Even if it was say 45 sec, how does the breakup of GET help make things better?

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -1358,87 +1366,209 @@ class AttributionService:
...
+
+        if result is None:
+            if self._remaining_decision_timeout() <= 0:
+                record_profiling_event(
+                    ProfilingEvent.ATTRIBUTION_GET_COMPLETED,
+                    node_id=node_id,
+                )
+                return False
```
hexinw-nvidia: Actually, why do we need to do another timeout check here after the _get_results() call? If we don't do the check here, we would have one extra rendezvous loop iteration on timeout. But the code is simpler.

I'd proposed we remove line 1443-1448

src/nvidia_resiliency_ext/shared_utils/health_check.py
```diff
@@ -1358,87 +1366,209 @@ class AttributionService:
...
+            return
+        self._ensure_decision_deadline()
+        timeout = min(_ATTRIBUTION_REQUEST_TIMEOUT_SECONDS, self._remaining_decision_timeout())
+        if timeout <= 0:
+            logger.info(
+                "AttributionService terminal POST skipped for %s: attribution decision "
+                "budget exhausted",
+                log_path,
+            )
+            return
```
hexinw-nvidia: Do we need the timeout check here? Seems to me it can be removed.

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
hexinw-nvidia: The GET(timeout=0.5) can return in milliseconds, not 0.5 seconds. Hence, we need the time.sleep(0.1). 

  if not self._attribution_gate(node_desc):
      time.sleep(0.1)   # avoid tight spin if attrsvc returns in_flight fast
      continue

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
hexinw-nvidia: Please remind me. I don't recall this is what we discussed. "return True" in line 1382 seems the correct implementation.

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
namitdhameja: There is no tight spin here. As mentioned, if the result is not available then it will wait for <budget>. I will make the change just because its not worth getting stuck on nit.

hexinw-nvidia: This is where the disconnect is. Why do we want the get_last_result() block if there is no result?

namitdhameja: I have already made the change you suggested:
if not self._attribution_gate(node_desc):
                time.sleep(0.1)
                continue
               
See the 0.1 sleep.
The behavior is for GET to wait till the result is available. The client can specify their own timeout if they don't want to wait.

hexinw-nvidia (approved):
LG. Thanks for addressing the comments.

PR #343, opened 2026-05-28
attribution_in_progress_end

helisha91:


src/nvidia_resiliency_ext/attribution/orchestration/config.py
```diff
@@ -25,8 +25,8 @@
 # LLM defaults — override with NVRX_LLM_MODEL / NVRX_LLM_BASE_URL env vars.
 # Default endpoint is build.nvidia.com (publicly accessible).
 # Internal NVIDIA users can override to inference.nvidia.com via NVRX_LLM_BASE_URL.
-DEFAULT_LLM_MODEL = os.environ.get("NVRX_LLM_MODEL", "nvidia/nemotron-3-super-120b-a12b")
-DEFAULT_LLM_BASE_URL = os.environ.get("NVRX_LLM_BASE_URL", "https://integrate.api.nvidia.com/v1")
+DEFAULT_LLM_MODEL = os.environ.get("NVRX_LLM_MODEL", "nvidia/nvidia/nemotron-3-super-v3")
```
hexinw-nvidia: Duplicated `nvidia/` prefix in the default model — `"nvidia/nvidia/nemotron-3-super-v3"`. Should be `"nvidia/nemotron-3-super-v3"`.

src/nvidia_resiliency_ext/attribution/orchestration/config.py
```diff
@@ -25,8 +25,8 @@
...
+DEFAULT_LLM_BASE_URL = os.environ.get("NVRX_LLM_BASE_URL", "https://inference-api.nvidia.com/v1")
```
hexinw-nvidia: This switches the default base URL from the public `integrate.api.nvidia.com` to the internal `inference-api.nvidia.com`, which contradicts the comment just above ("Default endpoint is build.nvidia.com (publicly accessible)... Internal users override via NVRX_LLM_BASE_URL"). This breaks the default for external users and looks like accidentally-committed internal test config. Revert unless intentional + update the comment.

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
```diff
@@ -369,8 +535,18 @@ def __init__(self, args: Union[argparse.Namespace, Mapping[str, Any]]):
...
+        self.job_inline_data_dict = {}
+        self.attribution_dict = {}
         self.exclude_nvrx_logs = bool(self._init_config.get("exclude_nvrx_logs", False))
         self.is_per_cycle = bool(self._init_config.get("is_per_cycle", False))
+        self.repeated_amount = (int(self._init_config.get("repeated_amount", 3)),)
+        self.stop_accumulating_count = (int(self._init_config.get("stop_accumulating_count", 3)),)
+        self.logs_minutes_before_job_end = (
+            int(self._init_config.get("logs_minutes_before_job_end", 20)),
+        )
+        self.chunks_per_time = (int(self._init_config.get("chunks_per_time", 5)),)
```
hexinw-nvidia: **Blocking:** trailing commas wrap each scalar in a 1-tuple — `self.repeated_amount` is `(3,)`, not `3`, etc. Every downstream use breaks at runtime:
- `time.sleep(self.chunks_per_time * 60)` → sleeps on a 300-element tuple → `TypeError`
- `empty_logs_stop -= 1` → `TypeError`
- `self.logs_minutes_before_job_end / self.chunks_per_time` → `TypeError`
- `self.cycle_counter_dict[...] == self.repeated_amount - 1` → `TypeError`

Drop the trailing commas.

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
```diff
@@ -401,6 +577,141 @@ async def analyze_logs(self) -> list[ApplicationData]:
...
+            if chunk_data.application_errors_list_full:
+                (
+                    application_log,
+                    attribution_raw_chunk,
+                    attribution_dict_chunk,
+                    hw_category_chunk,
+                ) = get_attribution(llm, app_data, True)
+                attribution_list.append(attribution_raw_chunk)
+
+            self.job_inline_data_dict[path].append(
```
hexinw-nvidia: **Blocking:** `job_inline_data_dict` is initialized as `{}` and `[path]` is never seeded to a list, so this `.append()` raises `KeyError` on the first poll. Add `self.job_inline_data_dict.setdefault(path, [])` before appending.

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
```diff
@@ -401,6 +577,141 @@ async def analyze_logs(self) -> list[ApplicationData]:
...
+                    attribution_dict_chunk,
+                    hw_category_chunk,
+                )
+            )
+            logger.info(
+                f"[ckpt] analyze_logs_rt_start poll #"
+                f"{len(self.job_inline_data_dict[path])}: "
+                f"chunk_data.checkpoint_saved="
+                f"{getattr(chunk_data, 'checkpoint_saved', False)}",
+                flush=True,
```
hexinw-nvidia: **Blocking:** `logging.Logger.info` does not accept a `flush` kwarg — this raises `TypeError: _log() got an unexpected keyword argument 'flush'`. Looks like a `print(flush=True)` → `logger.info` conversion. Remove `flush=True` here and at the other occurrences (lines 711, 919, 1013, 1022, 1043).

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
```diff
@@ -401,6 +577,141 @@ async def analyze_logs(self) -> list[ApplicationData]:
...
+            )
+
+            time.sleep(self.chunks_per_time * 60)
```
hexinw-nvidia: Blocking `time.sleep()` inside an `async def` blocks the event loop for the whole poll interval. Use `await asyncio.sleep(...)`.

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
```diff
@@ -401,6 +577,141 @@ async def analyze_logs(self) -> list[ApplicationData]:
...
+
+        return None
```
hexinw-nvidia: Return-type mismatch: `analyze_logs_rt_start` is annotated `-> list[ApplicationData]` but returns `None`. `ProgressiveLogAnalysisStartTool.run` then propagates this `None` while its unsupported branch returns a payload dict — downstream consumers expecting a payload may break on `None`.

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
```diff
@@ -553,6 +889,177 @@ async def llm_analyze(self, output_list: list[ApplicationData]) -> list[LogSageC
...
+            logger.info(
+                "[ckpt] _streaming_attribution: checkpoint_saved bypass FIRED "
+                "→ is_attribution_current_last forced to False; "
+                "repeated-issue STOP override will be skipped",
+                flush=True,
+            )
+
+        if (
+            is_attribution_current_last
+            and self.cycle_counter_dict[cycle_counter_key] == self.repeated_amount - 1
```
hexinw-nvidia: Two issues here: (1) `self.repeated_amount` is a 1-tuple due to the trailing-comma bug, so `self.repeated_amount - 1` raises `TypeError`. (2) `cycle_counter_dict[cycle_counter_key]` is only seeded when `cycle_counter == 0`; if a process first sees `cycle_counter > 0`, this access (and the `+= 1` below) raises `KeyError`. Use `.get(key, 0)` / `setdefault`.

hexinw-nvidia (commented):
Automated review of the progressive/streaming attribution path. Flagged 5 blocking runtime bugs (tuple-wrapped scalars, `flush=` on `logger.info`, uninitialized dict list, model typo, internal default base URL) plus a few correctness/test concerns inline. Paths #1-#3 would crash on first invocation, which suggests the new code was not exercised end-to-end; the added `run_*` files are manual drivers, not pytest-collected tests.

src/nvidia_resiliency_ext/attribution/orchestration/config.py
helisha91: @hexinw-nvidia 
I'm not familiar with public url. let me check it

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
helisha91: @hexinw-nvidia 
Screenshot 2026-05-31 at 14 51 16

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
helisha91: @hexinw-nvidia 
image

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
helisha91: @hexinw-nvidia 
removed
image

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
helisha91: @hexinw-nvidia 
image

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
helisha91: @hexinw-nvidia 
image

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
helisha91: @hexinw-nvidia 
image

src/nvidia_resiliency_ext/attribution/orchestration/config.py
namitdhameja, line 1: None of the changes in the file are needed, the internal endpoints are set via config. The external endpoint is the default.

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
namitdhameja, line 1: class LogSageRunner:
    """Runs LogSage once per path (lib or MCP), returning the MCP-shaped ``log_result`` dict."""

    def __init__(self, config: LogSageExecutionConfig):
        self.config = config
        self._lib_log_analyzer: Any = None
        # Set on first failed NVRxLogAnalyzer() so we do not retry init every request.
        self._lib_log_analyzer_init_error: Optional[BaseException] = None
        self._log_analysis_lock = asyncio.Lock()

>> YThe concern is that Logsage is already taking a lock here and then doing a wait in progressive read, this will block any further notification I think till that routinue finished.

namitdhameja (changes requested):

src/nvidia_resiliency_ext/attribution/orchestration/config.py
helisha91: Ok. I reverted the changes

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
helisha91: I will check it!

src/nvidia_resiliency_ext/attribution/orchestration/config.py
helisha91: @hexinw-nvidia 
reverted the changes

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
namitdhameja, line 1: The progressive history cache looks heavier than needed. job_inline_data_dict[path] grows by appending one tuple per poll, and those tuples retain raw lines, ApplicationData, and intermediate attribution objects. At final attribution time, we only seem to need the latest file offset, a bounded recent log window, and an accumulated checkpoint_saved flag.

Could we replace the append-only list with a single per-path progressive state object that reconciles each poll into that state, e.g. latest offset + rolling recent chunks + checkpoint_seen? Then after final attribution consumes it, we can drop that per-path state entirely. That would avoid unbounded memory growth across long-running services/jobs while preserving the behavior needed for final attribution.

namitdhameja: Note that a lock is needed, but probably not at logsage MCP level. analyze_logs_rt_start() appends while final analyze_logs() reads from the same dict. If start is still polling when final attribution begins, state can interleave.
Other option is progressive polling is moved to a background task, we could introduce a narrower per-path lock around the progressive state. The broad LogSageRunner lock should not be held while the poller sleeps, but the poller and final attribution still need synchronization so final attribution does not consume state while a poll is mutating it. A per-path state object with a lock, latest offset, bounded recent chunks, checkpoint_seen flag, and task/cancel state would make that lifecycle explicit. Final attribution can take the per-path lock, snapshot/reconcile the state, cancel or mark the poller closed, then remove the state after use.

@namitdhameja 
I added lock on these functions:
    def reconcile(self, file_offset: int, new_lines: list[str], checkpoint_saved: bool) -> None:
        """Fold one poll into the bounded state. Call under :meth:`lock`.

        Only *non-empty* polls enter ``recent_chunks``: empty reads (the writer
        has gone idle, but the start phase keeps polling until its deadline) must
        not evict real content from the bounded window — otherwise the last
        log chunk (which carries the terminal error) can scroll out before final
        attribution reads it. ``latest_offset`` / ``poll_count`` still advance on
        every poll.
        """
        if checkpoint_saved:
            if self.poll_count == 0:
                self.checkpoint_first_poll = True
            self.checkpoint_seen = True
        self.latest_offset = file_offset
        if new_lines:
            self.recent_chunks.append(list(new_lines))
        self.poll_count += 1

    def snapshot(self) -> "tuple[int, list[str], bool]":
        """Return ``(latest_offset, recent_lines, checkpoint_seen)`` atomically.

        Flattens the rolling window into a single list. Call under :meth:`lock`
        so a concurrent :meth:`reconcile` cannot observe a partial read.
        """
        lines: list[str] = []
        for chunk in self.recent_chunks:
            lines.extend(chunk)
        return self.latest_offset, lines, self.checkpoint_seen

    def request_close(self) -> None:
        """Signal the poller to stop; cancel its task if one is registered.

        Call under :meth:`lock` so the poller cannot reconcile after close.
        """
        self.closed = True
        task = self.poller_task
        if task is not None and not task.done():
            task.cancel()

namitdhameja (changes requested):

src/nvidia_resiliency_ext/attribution/log_analyzer/nvrx_logsage.py
helisha91: @namitdhameja 
You are right.
I fixed it in the last commit :)

hexinw-nvidia (approved):

namitdhameja (approved):

PR #345, opened 2026-06-01
fix(attribution): discover FR dumps beside app logs

namitdhameja:
## Summary
- keep `TORCH_FR_DUMP_TEMP_FILE=<prefix>` as the explicit first FR discovery rule when the prefix resolves under the attrsvc allowed root
- add fallback discovery for flat per-cycle logs where FR dumps live under `<log_dir>/checkpoints`, e.g. `/mnt/logs/test_job_cycle0.log` -> `/mnt/logs/checkpoints`
- retain the existing `<run>/checkpoints` fallback for logs under `<run>/logs/...`

## Root Cause
NVBUG 6233471 reports that launcher-managed attrsvc can produce app-log attribution while missing already-generated FR dumps. For flat per-cycle logs such as `/mnt/logs/test_job_cycle0.log`, the previous layout heuristic only tried `/mnt/checkpoints`, and the `TORCH_FR_DUMP_TEMP_FILE` fallback only helped when the analyzed app log contained a resolvable prefix. It did not try `/mnt/logs/checkpoints`.

## Validation
- `PYTHONPATH=src python3.11 -m unittest tests.attribution.unit.test_fr_dump_path -v`
- `python3 -m ruff check src/nvidia_resiliency_ext/attribution/trace_analyzer/fr_support.py tests/attribution/unit/test_fr_dump_path.py`
- Dev login repro against unpatched and patched worktrees:
  - unpatched: `resolved=None`, `pipeline_fr_analysis=None`
  - patched: resolves `.../mnt/logs/checkpoints`, FR analysis runs
- Dev Slurm repro job `3068564` on node `nvl72150-T07`: completed `0:0`; unpatched missed FR, patched resolved `.../mnt/logs/checkpoints` and produced `hanging ranks: [0]`


hexinw-nvidia:
**Thanks for this — the refactor is clean and the helpers de-duplicate the validation nicely. A couple of points before merge:**

**1. Resolution order changed — please call it out.** This flips the precedence: the log-scanned `TORCH_FR_DUMP_TEMP_FILE` prefix now wins *before* the inferred `<run>/checkpoints` dir (previously the inferred shared dir was preferred). Since `TORCH_FR_DUMP_TEMP_FILE` is often a container-local path, the only thing preventing a false match against an unrelated path that happens to exist on the attrsvc host (e.g. its own `/tmp/checkpoints/_dump_*`) is `allowed_root`. Worth an explicit note in the PR body and the docstring, since `allowed_root` is now the sole guard.

**2. Let's fix the root cause rather than keep growing the layout heuristics.** This PR brings the layout-guessing rules in `extract_fr_dump_path` to three (`TORCH_FR_DUMP_TEMP_FILE` scan, `<log_dir>/checkpoints`, `<run>/checkpoints`). Each new cluster convention tends to force another one — let's fix the root cause so we're not adding a fourth.

The underlying problem isn't the discovery algorithm; it's a missing data channel. attrsvc only receives `log_path` (`http_api.py:29-44` — no checkpoint/FR field), so it has to reverse-engineer the dump location from log scanning + filesystem-layout guessing.

The launcher already knows the answer: it submits via `_submit_log()` (`launcher.py:992-993`), and its `os.environ` contains `TORCH_FR_DUMP_TEMP_FILE` at submit time. Proposal:

- Add an optional `fr_dump_path` field to the submit payload (`log_submit_payload` / POST `/logs`).
- Have `ft_launcher` populate it from `os.environ["TORCH_FR_DUMP_TEMP_FILE"]` when present.
- In `extract_fr_dump_path`, prefer an explicitly-supplied `fr_dump_path` (still subject to the `allowed_root` + traces-exist validation), and fall back to the existing log-scan / layout heuristics only when it's absent (e.g. non-launcher submitters).

That makes resolution deterministic for launcher-managed jobs and demotes the heuristics to a genuine fallback.

**Suggestion:** I'm fine merging this as a stopgap if we open a tracking issue for the `fr_dump_path` change so we don't keep accreting layout rules. Happy to take that follow-up.

namitdhameja:
> **Thanks for this — the refactor is clean and the helpers de-duplicate the validation nicely. A couple of points before merge:**
> 
> **1. Resolution order changed — please call it out.** This flips the precedence: the log-scanned `TORCH_FR_DUMP_TEMP_FILE` prefix now wins _before_ the inferred `<run>/checkpoints` dir (previously the inferred shared dir was preferred). Since `TORCH_FR_DUMP_TEMP_FILE` is often a container-local path, the only thing preventing a false match against an unrelated path that happens to exist on the attrsvc host (e.g. its own `/tmp/checkpoints/_dump_*`) is `allowed_root`. Worth an explicit note in the PR body and the docstring, since `allowed_root` is now the sole guard.
> 
> **2. Let's fix the root cause rather than keep growing the layout heuristics.** This PR brings the layout-guessing rules in `extract_fr_dump_path` to three (`TORCH_FR_DUMP_TEMP_FILE` scan, `<log_dir>/checkpoints`, `<run>/checkpoints`). Each new cluster convention tends to force another one — let's fix the root cause so we're not adding a fourth.
> 
> The underlying problem isn't the discovery algorithm; it's a missing data channel. attrsvc only receives `log_path` (`http_api.py:29-44` — no checkpoint/FR field), so it has to reverse-engineer the dump location from log scanning + filesystem-layout guessing.
> 
> The launcher already knows the answer: it submits via `_submit_log()` (`launcher.py:992-993`), and its `os.environ` contains `TORCH_FR_DUMP_TEMP_FILE` at submit time. Proposal:
> 
> * Add an optional `fr_dump_path` field to the submit payload (`log_submit_payload` / POST `/logs`).
> * Have `ft_launcher` populate it from `os.environ["TORCH_FR_DUMP_TEMP_FILE"]` when present.
> * In `extract_fr_dump_path`, prefer an explicitly-supplied `fr_dump_path` (still subject to the `allowed_root` + traces-exist validation), and fall back to the existing log-scan / layout heuristics only when it's absent (e.g. non-launcher submitters).
> 
> That makes resolution deterministic for launcher-managed jobs and demotes the heuristics to a genuine fallback.
> 
> **Suggestion:** I'm fine merging this as a stopgap if we open a tracking issue for the `fr_dump_path` change so we don't keep accreting layout rules. Happy to take that follow-up.

@hexinw-nvidia The attrsvc gets the log_path but not the FR path. FR dump is completely controlled at the workload level, with no notification to attrsvc or any other outside component. https://docs.pytorch.org/tutorials/unstable/flight_recorder_tutorial.html#enabling-flight-recorder ; attrsvc gets to know the path via TORCH_FR_DUMP_TEMP_FILE being in the cycle log file. And if its not there, then its heuristics based. We can probably just do what we do for Megatron, and ignore other workloads. Lets check with bug submitter.

hexinw-nvidia (approved):

PR #353, opened 2026-06-23
Persist async checkpoint plans across resume

sbak5:
Store cached local save plans in serialized DCP planner metadata so the first save after restart can validate and reuse loaded global metadata.

Previously, NVRx attached the gathered local plans only as a dynamic `Metadata.all_local_plans` attribute. Newer PyTorch DCP writers may rebuild the `Metadata` dataclass before pickling `.metadata`, which drops dynamic attributes and prevents resumed jobs from finding `loaded_all_plans`.

Persist the plans under a reserved `Metadata.planner_data` key while preserving the legacy in-memory attribute for compatibility. Add a regression test that loads metadata back from disk and verifies a fresh metadata cache can recover the saved local plans.

### DIFF
--- a/src/nvidia_resiliency_ext/checkpointing/async_ckpt/state_dict_saver.py
+++ b/src/nvidia_resiliency_ext/checkpointing/async_ckpt/state_dict_saver.py
@@ -34,6 +34,8 @@ if TYPE_CHECKING:
 
 logger = getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


hexinw-nvidia (commented):
Have an offline discussion with @sbak5 to understand the change.
The change looks good overall. 
I will leave it to @shurkat-nvidia for the final approval. Thanks.

hexinw-nvidia:
Added "ci-approved" label to trigger the GitLab CI.

PR #355, opened 2026-06-24
[codex] Filter stale FR dumps by Pstart time

namitdhameja:
## Summary

Fixes the `ft_launcher` portion of [NVBug 6302348](https://nvbugspro.nvidia.com/bug/6302348): FR dump discovery can pick up stale `_dump_*` files from previous cycles when the same `TORCH_FR_DUMP_TEMP_FILE` prefix is reused.

This PR does not fully solve the service/splitlog case. In service splitlog mode, the submitted path can be a scheduler/slurm pointer file or log directory flow rather than the exact app log for a cycle, so there is not yet a reliable per-cycle Pstart timestamp to use as an FR lower-bound filter.

## Problem

`ft_launcher` runs repeated cycles and can reuse the same FR dump prefix across cycles. The FR analyzer currently discovers dump files by globbing the prefix, so stale files left behind from an older cycle can be included in the current cycle's analysis.

That can poison the result because FR analysis is then operating on a mixed set of dumps:

- old dump files from a previous cycle
- fresh dump files from the current cycle

For `ft_launcher`, the lifecycle gives us the useful timing anchor:

- `Pstart`: submit the exact per-cycle app log, for example `*_cycle<N>.log`
- `Pend`: terminal/end submit for that same cycle
- `Pget`: retrieve the result before the next cycle starts

The relevant lower bound is `Pstart`, not `Pend`.

## Fix

This PR records a wall-clock `submitted_at_wall` timestamp when a new non-terminal exact app-log job is created. During FR analysis, that timestamp is passed as `fr_min_mtime`, and FR dump discovery filters files with:

```text
dump_mtime >= Pstart
```

Direct per-cycle app logs matching `*_cycle<N>.log` are treated as exact app logs even when `job_id` is present, so the `ft_launcher` path does not get misclassified as pending splitlog/slurm mode.

Terminal-created jobs, pending jobs, and splitlog jobs leave `submitted_at_wall=None`, which preserves the existing no-filter behavior.

## Scope

Solved here:

- `ft_launcher` exact per-cycle app-log submissions
- stale FR dump files older than the cycle's `Pstart`
- lib FR analyzer path
- MCP FR analyzer path
- combined log+FR MCP path

Not solved here:

- service splitlog mode
- service submissions where the submitted path is only a slurm/pointer file and not the exact per-cycle app log
- per-cycle FR dump namespacing at the workload layer

For service mode, the current behavior remains intentionally conservative: if we cannot identify an exact per-cycle app-log Pstart, no mtime filter is generated.

## Validation

- `UV_CACHE_DIR=/private/tmp/uv-cache PYTHONPATH=src uv run --no-project --python 3.11 --with pytest python -m pytest -q tests/attribution/unit/test_progressive_plumbing.py tests/attribution/unit/test_log_analyzer_observability.py tests/attribution/unit/test_fr_mtime_filter.py tests/attribution/unit/test_fr_mcp_result.py`
- `UV_CACHE_DIR=/private/tmp/uv-cache PYTHONPATH=src uv run --no-project --python 3.11 --with pytest python -m pytest -q tests/attribution/unit/test_combined_log_fr.py tests/attribution/unit/test_mcp_server_cache.py tests/attribution/unit/test_fr_dump_path.py tests/attribution/unit/test_analysis_pipeline_trace_only.py`
- `black==24.10.0 black --check .`
- `isort==5.13.2 isort --check-only .`
- `ruff==0.8.3 ruff check .`
- `git diff --check`

### DIFF
--- a/src/nvidia_resiliency_ext/attribution/analyzer/engine.py
+++ b/src/nvidia_resiliency_ext/attribution/analyzer/engine.py
@@ -326,7 +326,12 @@ class Analyzer:
         except ValueError as e:
             return LogAnalyzerError(error_code=ErrorCode.INVALID_PARAMETER, message=str(e))
 
-        result = await self._log.submit(log_path, user=user, job_id=job_id)
+        result = await self._log.submit(
+            log_path,
+            user=user,
+            job_id=job_id,
+            record_submission_time=(intent != ANALYSIS_INTENT_TERMINAL),
+        )
         if isinstance(result, LogAnalyzerError):
             return result
 
--- a/src/nvidia_resiliency_ext/attribution/combined_log_fr/combined_log_fr_mcp.py
+++ b/src/nvidia_resiliency_ext/attribution/combined_log_fr/combined_log_fr_mcp.py
@@ -217,6 +217,8 @@ class CombinedLogFRMCPOrchestrator:
             "llm_analyze": bool(arguments.get("llm_analyze", False)),
             "threshold": arguments.get("threshold"),
         }
+        if arguments.get("fr_min_mtime") is not None:
+            fr_kw["fr_min_mtime"] = float(arguments["fr_min_mtime"])
 
         log_analyzer = NVRxLogAnalyzer(log_kw)
         fr_analyzer = CollectiveAnalyzer(fr_kw)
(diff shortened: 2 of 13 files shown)


sbak5:
Is there any use timestamp to distinguish FR traces? you can annotate file name or explicit deletion per cycle or create subdirectores with restart index for post-mortem  analysis?

PR #376, opened 2026-07-21
CPU Tensor snapshotting (clone) for async-shm 

ankurv-nvidia:
(Summary from bug: 6439229)

Checkpoints saved through the persistent async worker (AsyncCallsQueue(persistent=True)) can be torn / internally inconsistent: a checkpoint scheduled when AdamW optimizer state "step" was 1.0 was persisted with step=2.0, while the params and momenta in the same checkpoint are from step 1 - so a resumed run gets an optimizer step count that does not match the saved weights. This happens because CPU tensors are transferred to the worker by moving their storage into shared memory (the trainer's live tensor and the worker's copy alias the same buffer, with no snapshot), so an in-place mutation by the next optimizer.step() lands in checkpoint N whenever the worker has not yet serialized that tensor.

The minimal reproducer below hits it 5/5 runs on GB200; in the QA test case it is intermittent (timing dependent).

Evidence: sha256 of the stored bytes for optimizer.state.{1,3}.step matches float32 2.0 (d88c86f1...) while the value at schedule time was float32 1.0 (e00e5eb9...).

Mechanism (call chain):
1. PersistentAsyncCaller.schedule_async_call (async_ckpt/core.py) sends the AsyncRequest to the worker via a torch.multiprocessing JoinableQueue. ForkingPickler transfers CPU tensors by MOVING their storage into shared memory - after the put(), the trainer's live tensor and the worker's received tensor alias the same buffer (no snapshot).
2. The trainer blocks on preload_q.join() until the worker finishes preload_fn (FileSystemWriterAsync.preload_tensors).
3. preload_tensors snapshots GPU tensors via tensor.to("cpu"), but the non-CUDA branch appends the tensor BY REFERENCE without clone (filesystem_async.py, final loop in preload_tensors; the "already cloned if needed" comment only covers the view-clone in _clone_or_dequantize_if_needed - a 0-dim optimizer step tensor is not a view).
4. After the trainer resumes, the next optimizer.step() increments the AdamW "step" CPU scalar IN PLACE - the aliased shm buffer changes while the worker is still serializing checkpoint N -> the stored file gets the future value.

Notes:
- Affects any persistent-worker async save: both use_cpu_shm_for_gpu_tensors=True and the default GPU-IPC path pass uncached CPU tensors through the same queue path. The fork-based non-persistent path is immune (fork COW snapshots CPU memory).
- Impact is general: every stock torch optimizer keeps "step" as a 0-dim CPU tensor by default, so real training checkpoints can persist a step count inconsistent with the saved params/momenta. Any other live CPU tensor mutated in place after scheduling (custom counters, scalers) is equally affected.


src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -621,6 +624,21 @@ def combine_data(gpu_tensor_data, uncached_tensor_data, byte_io_data):
             storage_plan,
         ) = data_structure
 
+        if uncached_tensor_data is not None:
+            # Clone the cpu tensors so they are snapshotted properly
+            uncached_items, uncached_data = uncached_tensor_data
+            uncached_data = [
+                (
+                    data.clone()
```
shurkat-nvidia: `torch.clone` returns a tensor with a gradient. Should we use `detach().clone()` instead?

src/nvidia_resiliency_ext/checkpointing/utils.py
```diff
@@ -91,7 +91,11 @@ def preload_tensors(state_dict: Dict, non_blocking=True):
 
     def preload_tensor(in_var):
         if isinstance(in_var, torch.Tensor):
-            return in_var.detach().to("cpu", non_blocking=non_blocking)
+            in_var = in_var.detach()
+            if in_var.device.type == "cpu":
+                # Copy: a reference wouldnt snapshot the live tensor after the shm transfer to the worker
```
shurkat-nvidia: Typo, "wouldn't"

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -621,6 +624,21 @@ def combine_data(gpu_tensor_data, uncached_tensor_data, byte_io_data):
...
+                    if isinstance(data, torch.Tensor)
+                    and data.device.type == 'cpu'
+                    and data.untyped_storage().is_shared()
```
shurkat-nvidia: I'm confused by these conditions. If the tensor is already in shared storage, do we need to clone it?

src/nvidia_resiliency_ext/checkpointing/utils.py
```diff
@@ -91,7 +91,11 @@ def preload_tensors(state_dict: Dict, non_blocking=True):
...
+                return in_var.clone()
+            return in_var.to("cpu", non_blocking=non_blocking)
```
shurkat-nvidia: Is this redundant with the functionality in `FileSystemWriterAsync`?

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
ankurvdev: Yes. In fact we only need to deal with shared memory tensors because those are the only tensors at risk of their data could be overwritten by another iteration while we're writing to disk.
The non-shared memory-cpu tensors (in the worker process) are safe because their data cant be touched by the training process.

I'm not too sure how the `--use-persistent-ckpt-worker` works without `--async-ckpt-use-cpu-shm` since I dont fully understand how cpu tensors are snapshotted in that case between iterations. 

But i'll keep that out of scope for now since non `--async-ckpt-use-cpu-shm` seem to be broken with the test image i'm using

sbak5:
CPU tensors are copied before they're transferred. without shm option, they're cloned (copied). Pushing them to a worker doesn't lead to shared memory right away. pickle / unpickle happens for non-shm cpu tensors.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
sbak5, line 309: This is where the clone happens.

sbak5: Currently it clones when the storage is not contiguous (added previously to handle cases for this type of cpu tensor). If you find any thing which doesn't work due to changes in `state_dict`, you need to include the condition here.

ankurv-nvidia: Thanks for pointer

src/nvidia_resiliency_ext/checkpointing/utils.py
ankurv-nvidia: I believe claude added this to mimic the fix for TorchAsyncCheckpoint which is an alternative path to FileSystemWriterAsync

I'll revert this from the MR since FileSystemWriterAsync is being used everywhere

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
ankurv-nvidia: Thanks. This seems like a better place to capture the state. I'll change this to always clone for cpu for snapshotting otherwise we risk it being overwritten

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
ankurv-nvidia: I moved this as per suggestion from @sbak5

sbak5 (dismissed):

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -295,13 +298,10 @@ def _clone_or_dequantize_if_needed(ten: torch.Tensor):
...
                     ten = ten.dequantize()
-                    # GPU tensors will be moved to CPU in preload_tensors
                     return ten, True
-                # GPU tensors will be moved to CPU in preload_tensors
                 return ten, False
-            # For CPU tensors, clone if they are views to ensure contiguous storage
-            is_view = ten.untyped_storage().size() != ten.numel() * ten.itemsize
-            return (ten.clone() if is_view else ten), False
+            # No clone needed for use_cpu_shm_for_gpu_tensors=True. All tensors snapshotted to shm
+            return (ten if self.use_cpu_shm_for_gpu_tensors else ten.clone()), False
```
shurkat-nvidia: The logic here feels clunky, could be simplified, IMO.

```python
if ten.device.type == "cpu" and not self.use_cpu_shm_for_gpu_tensors:
    return ten.clone(), False
if ten.device.type == "cuda" and "dequantize" in type(ten).__dict__:
    return ten.dequantize(), True
return ten, False
```

The new logic does not account for the `is_view` condition either. Is this intentional?

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -317,62 +317,53 @@ def resolve_data(items):
...
                     uncached_data.append(data)
-                elif was_dequantized and not include_dequantized:
+                elif was_dequantized:
                     uncached_items.append(item)
                     uncached_data.append(data)
                 else:
-                    gpu_items.append(item)
-                    gpu_data.append(data)
+                    cacheable_items.append(item)
+                    cacheable_data.append(data)
```
shurkat-nvidia, line 335: Simplify?

```python
            for item, data, was_dequantized in zip(items, resolved_data, dequantized_flags):
                if (isinstance(data, torch.Tensor) and data.device.type == "cpu") or was_dequantized:
                    uncached_items.append(item)
                    uncached_data.append(data)
                else:
                    cacheable_items.append(item)
                    cacheable_data.append(data)
```

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -383,55 +374,65 @@ def separate_cacheable(items, resolved_data, dequantized_flags, include_dequanti
...
                         # Periodically sync to avoid exhausting CUDA's pageable staging
                         # pages (shm tensors are not pinned so copy_ is not a pinned DMA).
-                        if (i + 1) % 8 == 0:
+                        pending_cuda_copies += int(source_t.device.type == "cuda")
+                        if pending_cuda_copies == 8:
                             torch.cuda.synchronize()
-                    torch.cuda.synchronize()
+                            pending_cuda_copies = 0
+                    if pending_cuda_copies:
+                        torch.cuda.synchronize()
```
shurkat-nvidia, line 377: Not required to be fixed in this PR. Just for visibility/flagging.

`torch.cuda.synchronize()` is rather expensive. Why is this needed? The comment above doesn't sound very convincing. The D2H copy is the only thing, and it's happening on one stream. The CUDA driver should manage these requests.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -317,62 +317,53 @@ def resolve_data(items):
...
-                include_dequantized=self.use_cpu_shm_for_gpu_tensors,
-            )
+            if self.use_cpu_shm_for_gpu_tensors:
+                # The shared-memory path snapshots all(cpu+gpu) tensors together
+                cacheable_items, cacheable_data = tensor_items, resolved_tensors
+                uncached_items, uncached_data = [], []
+            else:
+                (cacheable_items, cacheable_data), (uncached_items, uncached_data) = (
+                    separate_for_gpu_ipc(tensor_items, resolved_tensors, dequantized_flags)
+                )
```
shurkat-nvidia:
```python
(cacheable_items, cacheable_data), (uncached_items, uncached_data) = (
    separate_for_gpu_ipc(tensor_items, resolved_tensors, dequantized_flags)
    if not self.use_cpu_shm_for_gpu_tensors
    else (tensor_items, resolved_tensors), ([], [])
)
```

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -383,55 +374,65 @@ def separate_cacheable(items, resolved_data, dequantized_flags, include_dequanti
...
+                    # values in. Allocations are cached only when structure caching is enabled.
                     shm_tensors = []
                     total_bytes = 0
-                    for i, t in enumerate(gpu_data):
-                        tc = t.contiguous()
-                        shm_t = torch.empty(tc.shape, dtype=tc.dtype).share_memory_()
-                        shm_t.copy_(tc, non_blocking=True)
+                    pending_cuda_copies = 0
+                    for tensor in cacheable_data:
+                        contiguous_tensor = tensor.contiguous()
```
shurkat-nvidia, line 389: `tensor.contiguous()` returns an in-memory tensor with the same data, essentially forcing a copy. Seems very strange to do one copy into a contiguous tensor and another into `shm_t`.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -383,55 +374,65 @@ def separate_cacheable(items, resolved_data, dequantized_flags, include_dequanti
...
-                    torch.cuda.synchronize()
+                            pending_cuda_copies = 0
+                    if pending_cuda_copies:
+                        torch.cuda.synchronize()
                     if self.use_cached_data_structure:
-                        FileSystemWriterAsync._shm_tensor_cache[key] = (gpu_items, shm_tensors)
+                        FileSystemWriterAsync._shm_tensor_cache[key] = (
+                            cacheable_items,
+                            shm_tensors,
+                        )
```
shurkat-nvidia, line 407: Same comment as the other one, `torch.cuda.synchronize()` feels like extra steps. We should investigate if this is strictly necessary, it's overwhelming the logic here.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
ankurvdev: yes. we now clone regardless of whether its a view or now.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
ankurvdev: Yes. The D2H is the only thing. 

I'm not sure I follow the CUDA Driver management logic 

The call chain after this does not interface with the CUDA device at all. 
how would a CUDA driver resolve the queued up operations and synchronize with the followup code that treats the cpu snapshot as finalized without a cuda.synchronize here. 

Perhaps i'm missing something.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
ankurvdev: I'm trying to maintain parity with the existing logic. 
I'm not sure of the explanation either but given thats whats exists today i'm not trying to change it

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
ankurvdev: i'll resolve this comment and use the other thread for the discussion on any insights you might have regarding why we can get rid of it

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
shurkat-nvidia: It's out of scope for this PR.

My concern here is with the repeated `torch.cuda.synchronize()`. The D2H copies are all on a single stream, so the `cudaMemcpyAsync` calls underneath should all be serialized. We shouldn't need to synchronize every eight copies. Finally, when all the copies are done, we should use `torch.cuda.stream.Event.synchronize()` to only synchronize on the stream.

ankurv-nvidia: #405

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
ankurv-nvidia: #405

shurkat-nvidia (dismissed):

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -317,62 +313,50 @@ def resolve_data(items):
...
-                dequantized_flags,
-                include_dequantized=self.use_cpu_shm_for_gpu_tensors,
+            (cacheable_items, cacheable_data), (uncached_items, uncached_data) = (
+                (
+                    separate_for_gpu_ipc(tensor_items, resolved_tensors, dequantized_flags)
+                    if not self.use_cpu_shm_for_gpu_tensors
+                    else (tensor_items, resolved_tensors)
+                ),
+                ([], []),
             )
```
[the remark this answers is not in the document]
shurkat-nvidia: Prefer option 2 over option 1.

shurkat-nvidia (approved):

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
sbak5: This code is made assuming the GPU is stalled during checkpoint. (Anyhow, this is true because PCI-E can't be multiplexed so you need a way to interleave traffic efficiently if you want this step to run in the background). 
CUDA runtime holds a specific number of pages by default and after that it falls in on-demand allocation of pages, which is very slow. (easily reproducible). 
Given that, all memcpy happens on the default stream and there's no active kernel in the GPU, this sync works fine. 

the interval is set randomly but this sync is anyhow needed.  using default stream and stream wait should perform better than this device sync. 
#405 may need to be merged.

PR #390, opened 2026-08-03
Category-based STOP/RESTART decisions in restart_agent (L1 category selection + L4 gate)

continue-revolution:
## Summary

Adds category-based STOP/RESTART decisions to `restart_agent`: a curated 38-category taxonomy that the L1 LLM selects from, and an L4 policy gate that consults the selection before falling back to the strict 6-condition immediate-stop check. Also includes an L0 primary-candidate fix and a set of L1 contract repairs that let honest-but-slightly-non-compliant model responses through.

Final evaluation on the 79-case labeled subset across the 4 usable model targets (updated labels):

| target | decision | anchor | cat_id | reasoning | median/case | total wall |
|---|---|---|---|---|---|---|
| **gpt** | **100.0%** | 96.2% | **96.2%** | **98.7%** | 51 s | 69 min |
| qwen397b | 97.5% | 96.2% | 87.3% | 97.4% | **20 s** | **39 min** |
| **gemini** | **100.0%** | 91.1% | 92.4% | 92.2% | 63 s | 88 min |
| nemotron | 94.9% | 89.9% | 87.3% | 96.1% | 47 s | 66 min |

Compared to a pre-Option-A deterministic baseline of 74.7% decision accuracy on the same 79 cases.

## What this PR contains

Six analyzer-only feature commits under `src/nvidia_resiliency_ext/attribution/restart_agent/`, plus lint follow-ups:

1. **Option A — L1 category selection + L4 category-based decision gate**
   - New `l1/categories.py` with a curated 38-entry taxonomy (`id`, `name`, `description`, `decision`, `failure_domain`, `retry_outlook`).
   - L1 response schema extended with an optional `category_selection { category_id, category_confidence, category_rationale }` field.
   - L4 policy consults it via `_category_selection_qualifies_action()` with a confidence gate (default 80); the new `category_confirmed_stop` basis fast-paths to `WORKLOAD_UNRECOVERABLE` when the pick has `decision=STOP` at or above the threshold. Fallthrough (below threshold, `category_id=0`, or `EXCLUDED`) preserves the pre-existing typed-claim / deterministic pipeline unchanged.

2. **Option A tuning**
   - Broadens category names/descriptions for cats 6 / 9 / 11 / 16.
   - Promotes short 1-line descriptions into the prompt list (+2.4 KB) so the model actually sees the discriminators.
   - Adds `category_confidence_threshold` override on `L4PolicyInput`, read from `NVRX_CATEGORY_CONFIDENCE_THRESHOLD` in `decision_pipeline`. Default remains 80.

3. **Carve out ModuleNotFoundError-on-Lustre boundary between cat 13 and 16** — surgical fix for a single edge case (b1-056 in the eval) where cat 16's broadened description was catching import failures on Lustre-backed code paths.

4. **L0: promote earliest terminal cascade over later `root_candidate` when the cascade precedes it** — `_select_primary_candidate` used to ignore `cascade_candidate` roles entirely, so when a job ended with a scheduler cancel (`root_candidate`) that followed an earlier NCCL watchdog CUDA-launch-failure (`cascade_candidate`) with no upstream root visible, the SLURM cancel won by default and all L1 models faithfully followed L0's pick. Fixed by promoting the earliest terminal cascade when it precedes the earliest root_candidate.

5. **L1 contract repairs** — 5 non-semantic in-place normalizers wired into `_parse_model_evidence` before strict validation:
   - `repair_biconditional_unknowns` — normalize `(value=unknown, status≠unknown)` or `(value≠unknown, status=unknown)` pairs to `unknown/unknown` (fixes nemotron's honest-but-nonconforming responses)
   - `repair_overlong_lists` — truncate `plausible_causes`, `missing_evidence`, `related_failures` to their contract caps rather than reject the response
   - `repair_invalid_evidence_supports` — drop invalid support-tag strings (e.g. nemotron using `related_failures` as a supports tag); drop the evidence item entirely if nothing valid remains
   - `repair_overlong_category_rationale` — truncate `category_selection.category_rationale` to the 400-char cap
   - `repair_schema_version` — normalize similar-but-wrong `schema_version` strings (e.g. gemini's `restart_agent_decision_evidence.v1` bleed-through) when the payload is otherwise structurally shaped like our schema
   - Also **relaxes one contract rule** in `model_evidence_contract_errors`: waive the "evidence must support `failure_domain` / `retry_outlook_without_workload_change`" citation requirement when the corresponding claim is `value=unknown / status=unknown`. A claim of unknown/unknown means "no informative evidence for this dimension", so requiring a citation would be self-contradictory. Non-unknown claims still require citation - the anti-hallucination guardrail is kept for any dimension the model actually commits to.

## Design principle

L1 typed claims (`retry_outlook_without_workload_change`, `failure_domain`) remain in place. The category selection is an **additional, orthogonal signal** — when the picker returns a category with `decision=STOP` at confidence ≥ threshold, L4 uses it as a fast path to `WORKLOAD_UNRECOVERABLE` before running the strict 6-condition immediate-stop gate. Confidence below threshold or `category_id=0` (no match) falls through to the existing typed-claim / deterministic logic. **No behavior change for cases without an L1 result.**

Empirically, in the final GPT run the category gate handled 78/79 cases (20 STOP + 58 RESTART) and the policy-engine fallback correctly decided the 1 remaining case where L1 was uncertain.

## Attempted-and-reverted: Pattern A taxonomy sharpening

An additional commit sharpening 4 STOP/RESTART category boundaries (cats 14/16, 18/13, 22/9, 27/25) was applied and then **reverted** after empirical evaluation:

- **qwen397b**: cleanly gained the 3 targeted case flips (b1-006/008/009 got cat 27, b1-029 flipped RESTART→STOP correctly). Decision 97.5%→98.7%, cat_id 87.3%→92.4%.
- **nemotron**: destabilized across the taxonomy. Netted 3 targeted gains (b1-020, b1-075, b1-076) but 5+ regressions on unrelated categories - cat 18 in particular started firing broadly. Decision 94.9%→91.1%, cat_id 87.3%→70.9%.

Net across the affected 2 models: -2 decisions, -9 cat_id hits. Reverted to preserve the stable baseline. The taxonomy sharpening is a real improvement in principle but nemotron over-latches onto "prefer over cat X" phrasing. Follow-up work should try a shorter / less imperative wording that qwen397b would still respect but nemotron won't over-weight.

## Known residual misses (context for reviewers)

None prevent shipping. Logged here as the residual after several rounds of iteration:

### Reasoning inaccuracy (root-cause narrative quality, judged by gemini-flash)

12 non-`same_root` grades across the 4 models. Every one is either a genuine hard-case, a model-side bias, or a known LLM-side failure - not addressable via analyzer or contract logic:

- **P1 - Ambiguous log (b1-048)**: TCPStore-during-GPTDataset-index-build. Two valid causal readings (rendezvous vs missing dataset files); the seed labeler themselves reclassified this case during human review.
- **P2 - Anchor-selection bias (b1-014/015 nemotron, b1-058/062 gemini)**: model anchors on downstream terminal event instead of earlier upstream signal. Model-side picker bias - would need L1 prompt work with unclear regression risk. Zero decision impact.
- **P3 - L1 fall-through to deterministic template (b1-031/061 gemini; b1-007 qwen397b)**: empty model output or provider HTTP 500. Not repairable via contract layer.
- **P4 - Narrative-picker mismatch (b1-056 nemotron)**: narrative describes a workload code error but typed claims land on the correct filesystem-visibility category.
- **P5 - Sparse log (b1-078 nemotron)**: 1-line SLURM cancel-due-to-requeue log. Model doesn't recognize the pattern from a single line.
- **P6 - Partial mechanism (b1-017 gpt)**: mechanism right, minor detail missing. Judge-side strictness rather than a real quality issue.

### Categorization inaccuracy (category-id mismatch vs GT)

29 miss-instances across the 4 models. Three patterns:

- **Pattern A - Taxonomy discrimination** (12 misses): same failure mechanism, sub-category ambiguity. Affects qwen397b (b1-006/007/008/009/029/055/056) and nemotron (b1-020/042/044/074/075/076). Fix attempted and reverted - see above. These stay in the residual.
- **Pattern B - Anchor-selection** (~10 misses): b1-058 hits gpt/qwen397b/gemini; b1-062 hits qwen397b/gemini/nemotron with each picking a different signal from an ambiguous log. Model-side picker bias.
- **Pattern C - No-match** (~7 misses): b1-003 nemotron (anchored on teardown, missed OOM kill), b1-031 gemini (rare cat 28 SDC), b1-048 gpt/qwen397b/gemini (ambiguous log), b1-061 gemini (sparse).

### Restart/stop inaccuracy (decision-accuracy misses)

6 misses total, all directly caused by picking the wrong side of a STOP/RESTART category boundary (Pattern A above):

| case | model | wrong pick | correct pick | direction |
|---|---|---|---|---|
| b1-029 | qwen397b | cat 13 (RESTART) | cat 18 (STOP) | called RESTART, should be STOP |
| b1-056 | qwen397b | cat 16 (STOP) | cat 13 (RESTART) | called STOP, should be RESTART |
| b1-020 | nemotron | cat 9 (RESTART) | cat 22 (STOP) | called RESTART, should be STOP |
| b1-074 | nemotron | cat 16 (STOP) | cat 14 (RESTART) | called STOP, should be RESTART |
| b1-075 | nemotron | cat 16 (STOP) | cat 14 (RESTART) | called STOP, should be RESTART |
| b1-076 | nemotron | cat 16 (STOP) | cat 14 (RESTART) | called STOP, should be RESTART |

## Contract-repair logic and design rationale

The 5 in-place repair functions + 1 relaxation were added incrementally in response to empirical failures. Design constraints:

- **All repairs are in-place mutations on the parsed payload before validation.** No retry, no re-prompt, no additional model call.
- **Every repair emits a note** (returned by the repair function) for observability. The current code doesn't thread the notes into the trace anomalies yet - infrastructure is in place for a follow-up.
- **Each repair is silent-forward but explicit-reverse**: on the way in, the modified payload is treated exactly like a valid one. If a reviewer wants to see what the model originally sent, the raw model output is preserved in `l1.raw_model_output` in the trace.
- **The contract relaxation for unknown claims** is a semantic change, not a repair. It changes what is *considered valid* rather than fixing an invalid payload. Rationale: the citation requirement for `failure_domain` / `retry_outlook` conflicts with an honest `unknown/unknown` claim - if the model correctly reports it has no informative evidence for a dimension, requiring a citation is self-contradictory. Non-unknown claims still require citation, so the anti-hallucination guardrail is kept for any dimension the model actually commits to.

## Test plan

- [ ] Reviewer: sanity-check `l1/categories.py` — 38 entries, each with all five fields populated
- [ ] Reviewer: L4 fallthrough — verify `l1_category_selection=None` produces the same rule/basis as before Option A on a case without L1
- [ ] Reviewer: env-var behavior — `NVRX_CATEGORY_CONFIDENCE_THRESHOLD` unset ⇒ default 80; `=70` ⇒ used by L4
- [ ] Reviewer: L1 response contract — `category_selection` is optional in the schema; existing L1 responses without it still validate
- [ ] Reviewer: repair chain in `l1/validation.py` — each `repair_*` function is a no-op on already-valid input and mutates only the specific misshapen field on input that matches its signature
- [ ] Reviewer: L0 change in `l0/assembly.py::_select_primary_candidate` — cases with a legitimate upstream `root_candidate` before any `cascade_candidate` still pick that root; the promotion path only fires when cascade precedes root

## Not in scope

- Evaluation harness — measured externally, kept out of this PR by design
- Anchor-selection improvements beyond the L0 cascade-promotion fix (Pattern B in the miss report). Zero decision impact under current gate.
- Multi-model threshold policy beyond the default 80 and the eval-harness GPT=70 shim
- Pattern A taxonomy sharpening beyond the 4 already-committed cat 6/9/11/16 broadenings — deferred pending a wording that works for nemotron
- CSV-backed taxonomy — a follow-up refactor to let QA push category revisions as a CSV drop
- Threading repair notes into trace anomalies for observability - follow-up

### DIFF
--- a/src/nvidia_resiliency_ext/attribution/restart_agent/decision_pipeline.py
+++ b/src/nvidia_resiliency_ext/attribution/restart_agent/decision_pipeline.py
@@ -122,6 +122,7 @@ def build_decision_outcome(
             assessment_grounded=l2_result.recovery_assessment_policy_grounded,
             retry_policy=RetryPolicyConfig.from_mapping(execution_context.retry_policy),
             declared_recovery_capabilities=(execution_context.declared_recovery_capabilities),
+            l1_category_selection=l1_result.category_selection(),
         )
     )
     primary = l4_outcome.primary
--- a/src/nvidia_resiliency_ext/attribution/restart_agent/decision_pipeline.py
+++ b/src/nvidia_resiliency_ext/attribution/restart_agent/decision_pipeline.py
@@ -5,8 +5,28 @@
 
 from __future__ import annotations
 
+import os
 from dataclasses import dataclass
 from typing import Any, Mapping
+
+
+def _category_threshold_override() -> int | None:
+    """Read per-model confidence threshold override from env var.
+
+    NVRX_CATEGORY_CONFIDENCE_THRESHOLD is set by the eval harness / caller
+    based on the model target (e.g. 70 for GPT, 80 default). Returns None
+    when unset to keep the module-level default (80).
+    """
+    raw = os.environ.get("NVRX_CATEGORY_CONFIDENCE_THRESHOLD")
+    if not raw:
+        return None
+    try:
+        val = int(raw.strip())
+    except ValueError:
+        return None
+    if 0 <= val <= 100:
+        return val
+    return None
 
 from .attempt_records import AttemptRecordAssembler
 from .causality import build_result_cascades
@@ -123,6 +143,7 @@ def build_decision_outcome(
             retry_policy=RetryPolicyConfig.from_mapping(execution_context.retry_policy),
             declared_recovery_capabilities=(execution_context.declared_recovery_capabilities),
             l1_category_selection=l1_result.category_selection(),
+            category_confidence_threshold=_category_threshold_override(),
         )
     )
     primary = l4_outcome.primary
--- a/src/nvidia_resiliency_ext/attribution/restart_agent/decision_pipeline.py
+++ b/src/nvidia_resiliency_ext/attribution/restart_agent/decision_pipeline.py
@@ -9,25 +9,6 @@ import os
 from dataclasses import dataclass
 from typing import Any, Mapping
 
-
-def _category_threshold_override() -> int | None:
-    """Read per-model confidence threshold override from env var.
-
-    NVRX_CATEGORY_CONFIDENCE_THRESHOLD is set by the eval harness / caller
-    based on the model target (e.g. 70 for GPT, 80 default). Returns None
-    when unset to keep the module-level default (80).
-    """
-    raw = os.environ.get("NVRX_CATEGORY_CONFIDENCE_THRESHOLD")
-    if not raw:
-        return None
-    try:
-        val = int(raw.strip())
-    except ValueError:
-        return None
-    if 0 <= val <= 100:
-        return val
-    return None
-
 from .attempt_records import AttemptRecordAssembler
 from .causality import build_result_cascades
 from .l1 import L1EvidenceResult, execution_status
@@ -51,6 +32,25 @@ from .models import (
     RetryPolicyConfig,
 )
 from .runtime import SYSTEM_CLOCK, Clock
+
+
+def _category_threshold_override() -> int | None:
+    """Read per-model confidence threshold override from env var.
+
+    NVRX_CATEGORY_CONFIDENCE_THRESHOLD is set by the eval harness / caller
+    based on the model target (e.g. 70 for GPT, 80 default). Returns None
+    when unset to keep the module-level default (80).
+    """
+    raw = os.environ.get("NVRX_CATEGORY_CONFIDENCE_THRESHOLD")
+    if not raw:
+        return None
+    try:
+        val = int(raw.strip())
+    except ValueError:
+        return None
+    if 0 <= val <= 100:
+        return val
+    return None
 
 
 @dataclass(frozen=True)
--- a/src/nvidia_resiliency_ext/attribution/restart_agent/l0/assembly.py
+++ b/src/nvidia_resiliency_ext/attribution/restart_agent/l0/assembly.py
@@ -1121,14 +1121,15 @@ def _selection_summary(
 
 def _line_numbering_anomaly() -> Mapping[str, str]:
     return {
-        "scheme": "linux_physical_lines",
+        "scheme": "python_text_universal_newlines",
         "line_field": (
-            "1-based physical line split on LF; a preceding CR is removed for "
-            "CRLF, while bare CR remains line content"
+            "1-based logical line after Python text-mode universal-newline "
+            "splitting on LF, CR, or CRLF"
         ),
         "shell_tool_note": (
-            "Line numbers align with Linux tools that count LF-delimited "
-            "records. Rank prefixes may also look like '<rank>:'."
+            "Rank prefixes also look like '<rank>:'. Tools that count only "
+            "LF-delimited records, or render embedded CR-delimited records, "
+            "may show different line numbers for the same text."
         ),
     }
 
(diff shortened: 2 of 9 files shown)


continue-revolution:
#411 is merged to replace this PR.

PR #398, opened 2026-08-08
Consume scheduler exclusions in FT rendezvous

namitdhameja:
## Summary

Consumes Scheduler Exclusion decisions in the FT rendezvous path.

- adds `--ft-scheduler-exclusion-dir` and its YAML equivalent
- adds `SegmentHealthCheck` for task- or segment-scoped health predicates
- lets only array-task Node0 read the compact first JSONL record
- validates exact task IDs, bounded input size, and artifact freshness while failing open on missing or invalid data
- raises the existing `UnhealthyNodeException` and reuses `unhealthy_replacement_groups_<round>`
- makes replacement-group identity generation-aware
- documents and provides the combined Slurm array lifecycle

Depends on #358, which owns the producer service and shared decision-artifact contract. After #358 merges, this PR can be retargeted to `main`.

## Validation

- Black, isort, shell syntax, and `git diff --check` passed
- focused FT, health-check, and scheduler-decision unit coverage is included
- prior PDX E2E covered baseline, exclusion without spares, replacement with a spare, stale/missing/malformed fail-open behavior, and typical decision-read overhead of about 2-3 ms

### DIFF
--- a/src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
+++ b/src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
@@ -71,6 +71,7 @@ from .cycle_info_writer import CycleInfoReporter, CycleInfoRoundSnapshot, cycle_
 from .data import WorkloadAction
 from .ipc_connector import IpcConnector
 from .launcher import FT_LAUNCHER_IPC_SOCKET, UnhealthyNodeException, get_node_health_check
+from .segment_health_check import get_segment_health_check
 
 # Conditionally import failure injector (only active when NVRX_INJECT_GPU_FAILURE is set)
 if os.environ.get("NVRX_INJECT_GPU_FAILURE"):
@@ -220,6 +221,8 @@ class RendezvousSettings:
         nproc_per_node:
             Number of processes per node (local_world_size). Used to restore local_world_size
             when a standby node becomes active.
+        segment_health_check_dir:
+            Shared directory containing segment health decision artifacts.
     """
 
     run_id: str
@@ -231,6 +234,7 @@ class RendezvousSettings:
     segment: Optional[int] = None
     replacement_group_size: Optional[int] = None
     nproc_per_node: int = 1  # Default to 1 if not specified
+    segment_health_check_dir: Optional[str] = None
 
 
 @dataclass(eq=True, order=True, frozen=True)
@@ -2346,6 +2350,7 @@ class FtRendezvousBarrierHandler(RendezvousHandler):
         attribution_enforce_stop: bool = False,
         cycle_info_dir: Optional[str] = None,
         cycle_log_prefix: Optional[str] = None,
+        segment_health_check_dir: Optional[str] = None,
     ):
         """Create a new :py:class:`FtRendezvousBarrierHandler`.
 
@@ -2396,6 +2401,7 @@ class FtRendezvousBarrierHandler(RendezvousHandler):
             segment=segment,
             replacement_group_size=replacement_group_size,
             nproc_per_node=nproc_per_node,
+            segment_health_check_dir=segment_health_check_dir,
         )
 
         return cls(
@@ -2451,6 +2457,10 @@ class FtRendezvousBarrierHandler(RendezvousHandler):
             )
         if settings.replacement_group_size is not None and settings.replacement_group_size < 1:
             raise ValueError("replacement_group_size must be a positive integer when configured.")
+        if settings.segment_health_check_dir and not os.path.isabs(
+            settings.segment_health_check_dir
+        ):
+            raise ValueError("segment_health_check_dir must be an absolute path.")
 
         self._this_node = node
         self._settings = settings
@@ -2500,6 +2510,7 @@ class FtRendezvousBarrierHandler(RendezvousHandler):
         self._storage_path_checker = (
             StoragePathHealthCheck(storage_healthcheck_paths) if storage_healthcheck_paths else None
         )
+        self._segment_health_checker = get_segment_health_check(settings.segment_health_check_dir)
 
         # Attribution service client (optional, only on master node)
         if is_store_host and attribution_endpoint:
@@ -2625,10 +2636,17 @@ class FtRendezvousBarrierHandler(RendezvousHandler):
             raise UnhealthyNodeException(str(e)) from e
 
     def ensure_node_is_healthy(self) -> None:
-        """Perform GPU, NIC link state, and Node health checks for this node."""
-        # Record the health check message
+        """Perform configured segment- and node-level health checks in order."""
+        # Record the health check before any configured check can raise.
         msg = f"Checking health status of {self._this_node}."
         self._record(message=msg)
+
+        if self._segment_health_checker is not None:
+            self._run_health_check(
+                self._segment_health_checker,
+                "Segment health check",
+                "The current Slurm array task segment is excluded by scheduler state.",
+            )
 
         # Perform GPU health check
         self._run_health_check(
@@ -3088,6 +3106,7 @@ def create_handler(
         )
         cycle_info_dir = params.config.get('cycle_info_dir', None)
         cycle_log_prefix = params.config.get('cycle_log_prefix', None)
+        segment_health_check_dir = params.config.get('segment_health_check_dir', None)
 
         return FtRendezvousBarrierHandler.from_backend(
             params.run_id,
@@ -3109,6 +3128,7 @@ def create_handler(
             attribution_enforce_stop=attribution_enforce_stop,
             cycle_info_dir=cycle_info_dir,
             cycle_log_prefix=cycle_log_prefix,
+            segment_health_check_dir=segment_health_check_dir,
         )
     except Exception as e:
         construct_and_record_rdzv_event(
--- a/src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
+++ b/src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
@@ -1,0 +1,65 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Consume segment health decisions before joining rendezvous."""
+
+from __future__ import annotations
+
+import logging
+import os
+import stat
+from pathlib import Path
+from typing import Optional
+
+from nvidia_resiliency_ext.shared_utils.log_manager import LogConfig
+
+log = logging.getLogger(LogConfig.name)
+
+
+class SegmentHealthCheck:
+    """Check whether a Slurm allocation unit remains eligible for rendezvous."""
+
+    def __init__(self, directory: str, job_id: str, task_id: str) -> None:
+        self.directory = directory
+        self.job_id = job_id
+        self.task_id = task_id
+        self._path = Path(directory) / f"segment_health_check.{job_id}.{task_id}"
+
+    def __call__(self) -> bool:
+        return self._perform_health_check()
+
+    def _perform_health_check(self) -> bool:
+        """Evaluate the current segment health artifact, failing open."""
+        try:
+            artifact = self._path.stat()
+        except FileNotFoundError:
+            return True
+        except OSError as exc:
+            log.warning("Ignoring segment health decision: %s", exc)
+            return True
+
+        if not stat.S_ISREG(artifact.st_mode):
+            log.warning("Ignoring segment health decision: not a regular file: %s", self._path)
+            return True
+
+        return artifact.st_size == 0
+
+
+def get_segment_health_check(directory: Optional[str]) -> Optional[SegmentHealthCheck]:
+    """Return the segment health check for this launcher, when eligible."""
+    if not directory or os.environ.get("SLURM_PROCID") != "0":
+        return None
+
+    job_id = os.environ.get("SLURM_ARRAY_JOB_ID") or os.environ.get("SLURM_JOB_ID")
+    if not job_id:
+        return None
+
+    task_id = os.environ.get("SLURM_ARRAY_TASK_ID") or job_id
+
+    log.info(
+        "Segment health check installed directory=%s job_id=%s task_id=%s",
+        directory,
+        job_id,
+        task_id,
+    )
+    return SegmentHealthCheck(directory, job_id, task_id)
--- a/docs/design/segment_health_check_ft_spec.md
+++ b/docs/design/segment_health_check_ft_spec.md
@@ -1,0 +1,73 @@
+# Segment Health Check: FT Consumer Design
+
+## Purpose
+
+NVRx consumes a shared segment health decision before joining rendezvous.
+The restart path performs filesystem I/O only; it does not query Slurm or call
+an external service. The
+[integration guide](../source/fault_tolerance/integration/segment_health_check.rst)
+owns the user-facing configuration, artifact contract, and supported launch
+shape.
+
+## Internal Flow
+
+```text
+FtRendezvousBarrierHandler initialization
+  -> get_segment_health_check()
+       -> ineligible launcher: no check installed
+       `-> eligible allocation-unit process 0: SegmentHealthCheck(directory, job_id, task_id)
+
+pre_join_hook
+  -> ensure_node_is_healthy()
+       -> SegmentHealthCheck._perform_health_check()
+       -> existing local node checks
+       `-> _run_health_check() converts failure to UnhealthyNodeException
+            -> array path marks replacement group unhealthy
+            `-> regular-job path exits the workload step
+```
+
+### Installation Guard
+
+`get_segment_health_check()` installs the check only on the launcher with
+`SLURM_PROCID=0`. Arrays use `SLURM_ARRAY_JOB_ID` and `SLURM_ARRAY_TASK_ID`;
+regular jobs use `SLURM_JOB_ID` as both the artifact scope and exclusion token.
+Missing job identity leaves the check uninstalled; other launchers return
+without logging. Successful installation emits one INFO record.
+
+### Decision Evaluation
+
+The concrete `SegmentHealthCheck` in
+`fault_tolerance/segment_health_check.py` owns the configured directory and
+Slurm job/task IDs. Its `_perform_health_check()`:
+
+1. derives `segment_health_check.<job_id>.<task_id>` from the configured
+   directory;
+2. performs one metadata check; and
+3. returns unhealthy only when the path is a non-empty regular file.
+
+The file content is producer-owned diagnostic context and is not read by the
+consumer. A missing or zero-byte file is healthy and quiet. An unreadable path
+or unexpected file type logs a warning and fails open.
+
+### Rendezvous Handoff
+
+A matching allocation ID returns unhealthy through `_run_health_check()`, which
+raises `UnhealthyNodeException`. For arrays, the existing pre-join exception
+path uses `SLURM_ARRAY_TASK_ID` as the replacement-group token and adds it to
+`unhealthy_replacement_groups_<round>`. A regular job has no replacement-group
+token; process 0 exits nonzero and `srun --kill-on-bad-exit=1` terminates its
+workload step. Segment Health Check adds no host-side selection path.
+
+## Implementation Map
+
+| Area | Responsibility |
+| --- | --- |
+| `fault_tolerance/segment_health_check.py` | Identify eligible process-0 launchers and implement `SegmentHealthCheck`. |
+| `fault_tolerance/config.py` and launcher plumbing | Validate and pass the decision directory into rendezvous settings. |
+| `fault_tolerance/ft_rendezvous_barrier.py` | Install the check and reuse the existing health-check exception path. |
+| `tests/fault_tolerance/unit/` | Cover installation, per-task file state, failure handling, and rendezvous propagation. |
+
+## Verification
+
+Environment-specific E2E scenarios, commands, and result analysis live with
+the deployment validation harness rather than this product design.
--- a/docs/source/fault_tolerance/integration.rst
+++ b/docs/source/fault_tolerance/integration.rst
@@ -9,4 +9,4 @@ Integration Guides
    integration/sections
    integration/ptl
    integration/inprocess
-
+   integration/segment_health_check
--- a/docs/source/fault_tolerance/integration/segment_health_check.rst
+++ b/docs/source/fault_tolerance/integration/segment_health_check.rst
@@ -1,0 +1,91 @@
+Segment Health Check
+====================
+
+The segment health check prevents NVRx from reusing a Slurm allocation unit
+after it is marked unavailable. The unit is an array task for a job array or
+the job itself for a regular Slurm job. The FT consumer reads a decision
+artifact from shared storage before joining rendezvous; it does not require
+Slurm access inside the workload environment. The decision publisher is
+independent of this consumer as long as it satisfies the artifact contract
+below.
+
+Configuration
+-------------
+
+Pass the shared decision directory to ``ft_launcher``:
+
+.. code-block:: bash
+
+   ft_launcher \
+     --ft-segment-health-check-dir /shared/job/segment-health-check \
+     <other-options> train.py
+
+The path must be absolute and accessible to every launcher. Omitting the option
+disables consumption. The equivalent YAML field is
+``fault_tolerance.segment_health_check_dir``.
+
+Supported Launch Shape
+----------------------
+
+The FT consumer supports Slurm job arrays and regular jobs. Both require one
+``ft_launcher`` process per allocated node and ``srun --kill-on-bad-exit=1``, so
+an excluded process 0 terminates the workload step. Job arrays additionally
+require ``--no-requeue`` for the initial deployment artifact lifecycle.
+
+Only the launcher with ``SLURM_PROCID=0`` in each allocation unit reads the
+artifact.
+
+Artifact Contract
+-----------------
+
+The consumer derives the following path in the configured directory:
+
+.. code-block:: text
+
+   segment_health_check.<job_id>.<task_id>
+
+For an array, ``job_id`` is ``SLURM_ARRAY_JOB_ID`` and ``task_id`` is
+``SLURM_ARRAY_TASK_ID``. For a regular job, both values are ``SLURM_JOB_ID``.
+
+A non-empty regular file marks the allocation unit unhealthy. A missing or
+zero-byte file proceeds normally. The file content is diagnostic context owned
+by the producer; the consumer performs one metadata check and does not read or
+parse it. Unreadable paths and unexpected file types emit a warning and fail
+open.
+
+Runtime Behavior
+----------------
+
+Only allocation-unit process 0 reads the artifact before rendezvous. A clean or
+fail-open result joins normally. A match raises ``UnhealthyNodeException``
+through the existing health-check path and exits nonzero. For an array, the
+same path also marks its replacement group unhealthy; the rendezvous host
+already ignores unhealthy replacement groups. A regular job has no array-task
+replacement group, so process-0 failure terminates its workload step through
+``srun --kill-on-bad-exit=1``.
+
+Other participants from that array task may join before process 0 publishes the
+unhealthy marker. Slurm then terminates the task after process 0 exits, which
+can require one additional rendezvous before eligible spares are selected.
+
+With enough eligible spares, rendezvous proceeds without the excluded task. If
+quorum cannot be formed, the existing rendezvous timeout applies; NVRx does not
+weaken the exclusion to create quorum.
+
+A scheduler state such as Slurm ``DRAIN`` does not evict a task that is already
+running. NVRx consumes the corresponding decision at the next rendezvous
+health-check boundary.
+
+Observability
+-------------
+
+An eligible allocation-unit process 0 logs once when the check is installed:
+
+.. code-block:: text
+
+   Segment health check installed directory=/shared/nvrx job_id=123 task_id=7
+
+A missing decision fails open quietly because the producer may not be running.
+An unreadable path or unexpected file type emits a warning. An exclusion is
+reported by the existing health-check error path. Other launchers do not
+install the segment check or inspect the artifact.
--- a/docs/source/fault_tolerance/usage_guide.rst
+++ b/docs/source/fault_tolerance/usage_guide.rst
@@ -199,6 +199,17 @@ Validation behavior:
   - Other existing types (e.g., devices/symlinks): performs ``stat`` access
 
 
+Segment Health Check
+^^^^^^^^^^^^^^^^^^^^
+
+Set ``--ft-segment-health-check-dir <ABSOLUTE_PATH>`` or
+``fault_tolerance.segment_health_check_dir`` to consume segment health
+decisions before rendezvous. Omitting the setting disables the check.
+
+See the :doc:`Segment Health Check integration guide
+<integration/segment_health_check>` for the consumer contract.
+
+
 Attribution service integration
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/src/nvidia_resiliency_ext/fault_tolerance/config.py
+++ b/src/nvidia_resiliency_ext/fault_tolerance/config.py
@@ -107,6 +107,9 @@ class FaultToleranceConfig:
       the cycle_info.<job_id>.current symlink there. The workload receives the path
       to the current cycle file via NVRX_CURRENT_CYCLE_INFO.
       Default: None (disabled).
+    * `segment_health_check_dir` [str|None] Shared directory containing per-task
+      segment health decision artifacts. Allocation-unit process 0 checks its artifact
+      before joining a rendezvous. Default: None (disabled).
 
     If any timeout is None, it has no effect (as if it was +INF).
     All timeouts can be deduced and set during runtime.
@@ -155,6 +158,9 @@ class FaultToleranceConfig:
 
     # NVRx cycle info: base directory for cycle_info JSON files
     cycle_info_dir: Optional[str] = None
+
+    # Segment health check: shared directory containing per-task decision artifacts
+    segment_health_check_dir: Optional[str] = None
 
     @property
     def is_progress_tracking_enabled(self) -> bool:
--- a/src/nvidia_resiliency_ext/fault_tolerance/launcher.py
+++ b/src/nvidia_resiliency_ext/fault_tolerance/launcher.py
@@ -2556,6 +2556,16 @@ def get_args_parser() -> ArgumentParser:
     )
 
     parser.add_argument(
+        "--ft-segment-health-check-dir",
+        "--ft_segment_health_check_dir",
+        type=str,
+        default=None,
+        dest="ft_segment_health_check_dir",
+        help="Shared directory containing per-task decisions consumed by the segment health check. "
+        "The feature is disabled when omitted.",
+    )
+
+    parser.add_argument(
         "--ft-skip-section-response",
         "--ft-skip_section_response",
         type=lambda x: str(x).lower() in ["true", "1", "yes"],
@@ -2960,6 +2970,14 @@ def config_from_args(args, launcher_pipe_read_fd=None, launcher_log_file=None) -
 
     fault_tol_cfg = FaultToleranceConfig.from_args(args)
     _validate_attribution_requires_per_cycle_applog(args, fault_tol_cfg)
+
+    if fault_tol_cfg.segment_health_check_dir:
+        if not os.path.isabs(fault_tol_cfg.segment_health_check_dir):
+            raise ValueError(
+                "--ft-segment-health-check-dir must be an absolute path, got: "
+                f"{fault_tol_cfg.segment_health_check_dir}"
+            )
+        rdzv_configs['segment_health_check_dir'] = fault_tol_cfg.segment_health_check_dir
 
     # Pass segment-related configs to rendezvous config
     rdzv_configs['segment'] = fault_tol_cfg.segment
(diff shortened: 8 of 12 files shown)


src/nvidia_resiliency_ext/fault_tolerance/scheduler_exclusion.py
```diff
@@ -0,0 +1,156 @@
...
+    slurm_array_job_id_from_env,
+    slurm_array_task_id_from_env,
+)
+from nvidia_resiliency_ext.shared_utils.log_manager import LogConfig
+
+log = logging.getLogger(LogConfig.name)
+
+_MAX_DECISION_LINE_BYTES = 8 * 1024
+_DECISION_MAX_AGE_SECONDS = 30 * 60
+_COMPACT_DECISION_PATTERN = re.compile(rb'\[(?:"[0-9]+"(?:,"[0-9]+")*)?\]')
```
hexinw-nvidia: **Blocker** (with lines 57 and 64).

`_COMPACT_DECISION_PATTERN` makes the contract byte-exact. I ran six plausible encodings of the *same* JSON document through the consumer:

```
EXCLUDED                    ["7","12"]          <- the contract
fails open (NOT excluded)   ["7", "12"]         <- json.dumps() default separators
fails open (NOT excluded)   ["7","12"]\r\n      <- CRLF
fails open (NOT excluded)   pretty-printed
fails open (NOT excluded)   [7,12]              <- ints, not strings
fails open (NOT excluded)   schema record first
fails open (NOT excluded)   trailing space
```

The integration guide says the publisher is independent "as long as it satisfies the artifact contract below", then shows `["7","12"]` without stating that whitespace, int-vs-string and line endings are load-bearing. Anyone writing a second producer from that doc reaches for `json.dumps(ids)` and gets a feature that logs `outcome=invalid` and waves every task through.

`json.loads` + `task_id in decision` is compatible with the current producer, tolerant of all six rows above, and forward-compatible with an object-shaped artifact if #399 adds node evidence (`in` works the same on list and dict). It also removes the "7 vs 17" token reasoning and `UnicodeEncodeError` from the except clause: ~9 lines -> ~4. The stated motivation — avoiding a decoded list — is not a real cost for a <=8 KiB read once per rendezvous.

src/nvidia_resiliency_ext/fault_tolerance/scheduler_exclusion.py
```diff
@@ -0,0 +1,156 @@
...
+
+            # The first record contains only quoted task IDs, so a fixed token cannot
+            # confuse task "7" with task "17" and avoids allocating a decoded list.
+            task_token = b'"' + self.task.task_id.encode("ascii") + b'"'
+            if task_token in decision:
+                outcome = "excluded"
+                return False
+            return True
+        except FileNotFoundError:
+            outcome = "missing"
```
hexinw-nvidia: **Blocker.**

`missing` logs at DEBUG, and there is no log at all confirming the check was installed. Combined with the encoding failures above and the group-size guard at line 122, this feature has three independent silent-off paths and no positive signal.

Two small additions make it QA-able: one INFO at install time (`Scheduler Exclusion armed: dir=... job_id=... task_id=...`), and raise `missing`/`expired` to WARNING when the directory was *explicitly configured* — a configured-but-absent artifact is an operational fault, not a normal state.

src/nvidia_resiliency_ext/fault_tolerance/scheduler_exclusion.py
```diff
@@ -0,0 +1,156 @@
...
+            raw_node_count,
+        )
+        return None
+
+    return SlurmArrayTask(job_id, task_id)
+
+
+def _decision_file_path(directory: str, job_id: str) -> Path:
+    base = Path(directory)
+    if not base.is_absolute():
```
hexinw-nvidia: Please drop this check.

Third absolute-path validation: already done in `launcher.config_from_args` and again in `FtRendezvousBarrierHandler.__init__`. Since the handler validates before it constructs `SegmentHealthCheck`, this one is unreachable through supported config — and it is the worst of the three, because it turns a config error into a silent per-round fail-open rather than a startup failure.

`test_missing_decision_and_relative_directory_fail_open` is covering a path that cannot occur.

src/nvidia_resiliency_ext/fault_tolerance/scheduler_exclusion.py
```diff
@@ -0,0 +1,156 @@
...
+    job_id = slurm_array_job_id_from_env(environ)
+    task_id = slurm_array_task_id_from_env(environ)
+    if not job_id or not task_id:
+        log.warning(
+            "Scheduler Exclusion is not installed because array-task Node0 is missing "
+            "SLURM_ARRAY_JOB_ID or SLURM_ARRAY_TASK_ID"
+        )
+        return None
+
+    raw_node_count = environ.get("SLURM_NNODES") or environ.get("SLURM_JOB_NUM_NODES")
```
hexinw-nvidia: Question.

The gate `replacement_group_size == SLURM_NNODES` decides whether the check installs at all, and failing it disables the feature with a single warning. Inside an srun step, whether `SLURM_NNODES` reports the job's or the step's node count varies by Slurm version, and `nvrx_singleton_array.sbatch` runs more than one srun step per task.

Is `SLURM_JOB_NUM_NODES` (job-scoped by definition) the safer primary here, with `SLURM_NNODES` as fallback rather than the other way round? Either way this needs verifying against the real deployment, since getting it wrong is silently-off rather than loudly-wrong.

src/nvidia_resiliency_ext/fault_tolerance/scheduler_exclusion.py
```diff
@@ -0,0 +1,156 @@
...
+        return self._perform_health_check()
+
+    def _perform_health_check(self) -> bool:
+        """Evaluate the current Scheduler Exclusion artifact, failing open."""
+        check_started = time.monotonic()
+        outcome = "not_excluded"
+        try:
+            decision_path = _decision_file_path(self.directory, self.task.job_id)
+            decision, modified_at = _read_first_control_record(decision_path)
+            if modified_at + _DECISION_MAX_AGE_SECONDS <= time.time():
```
hexinw-nvidia: Documentation request.

This 30-minute bound is measured from the artifact's mtime, which makes "the producer republishes every pass even when unchanged" a load-bearing part of the contract. That invariant currently lives only in a comment in #399's header.

Please state it in the integration guide's Artifact Contract section, and note the coupling: `POLL_INTERVAL` (600s) must stay comfortably under this value, and neither side validates the other.

src/nvidia_resiliency_ext/fault_tolerance/scheduler_exclusion.py
```diff
@@ -0,0 +1,156 @@
...
+                "Scheduler Exclusion FT check job_id=%s task_id=%s outcome=%s elapsed_ms=%.3f",
+                self.task.job_id,
+                self.task.task_id,
+                outcome,
+                elapsed_ms,
+            )
+        return True
+
+
+def is_current_array_task_node0(env: Optional[Mapping[str, str]] = None) -> bool:
```
hexinw-nvidia: Nit: `is_current_array_task_node0` has one caller, which then re-reads `SLURM_ARRAY_TASK_ID` immediately. Worth folding into `scheduler_exclusion_task_for_launcher`.

src/nvidia_resiliency_ext/shared_utils/job_metadata.py
```diff
@@ -17,6 +17,16 @@
 """Helpers for job metadata derived from launcher environment variables."""
 
 import os
+from typing import Mapping
+
+
+def _nonempty_env_value(name: str, env: Mapping[str, str] | None = None) -> str | None:
```
hexinw-nvidia: Please drop the `job_metadata.py` change entirely.

These three helpers have exactly three call sites, all in `scheduler_exclusion.py`, all inside functions that already receive an `environ` mapping. The body is `environ.get(name, "").strip() or None`.

They also leave this module with two inconsistent readers of `SLURM_ARRAY_JOB_ID`: `job_id_from_env()` does not strip and takes no `env` parameter, these do both.

Suggest dropping this file and `tests/shared_utils/test_job_metadata.py` from the PR and inlining one private helper in `scheduler_exclusion.py`. Smaller blast radius — no shared-module change in a feature PR.

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -2625,7 +2646,14 @@ def _run_health_check(self, health_checker, check_name: str, failure_message: st
             raise UnhealthyNodeException(str(e)) from e
 
     def ensure_node_is_healthy(self) -> None:
-        """Perform GPU, NIC link state, and Node health checks for this node."""
+        """Perform configured segment- and node-level health checks in order."""
+        if self._segment_health_checker is not None:
```
hexinw-nvidia, line 2644: Nit: the segment check runs before `self._record(msg)`, so an excluded node emits no rendezvous health-check event and disappears from the event stream without a trace. Deliberate ordering is fine — consider recording first, or emitting a dedicated record for the exclusion.

hexinw-nvidia (commented):
The rendezvous integration is the right shape — reusing `_run_health_check` and the existing `UnhealthyNodeException` path means no new host-side selection logic, and the install guard correctly limits reads to array-task Node0.

One blocker on the artifact parse, one on observability. Both are about the same underlying property: this is a safety mechanism with three independent paths that disable it silently, and today nothing distinguishes "working" from "off". That matters more than usual for QA — a run where the poller was never started is indistinguishable from a clean run.

Note the base branch is `ft-node-state-service` (#358), which is now deprecated. Please rebase onto `main` so this can be reviewed and merged against what actually ships.

Two non-blocking points that do not anchor to a line:

- Neither this PR nor #399 wires the feature into `nvrx_singleton_array.sbatch` — no `--ft-scheduler-exclusion-dir`, no poller launch. It ships documented but not runnable end-to-end, which makes QA more expensive rather than less. Worth adding to one of the two before validation.
- Worth a line in the guide for QA's benefit: draining a node does **not** evict a running task; it only prevents rejoin at the next rendezvous. Someone testing this will drain a node, see nothing happen, and report it as broken.

src/nvidia_resiliency_ext/fault_tolerance/scheduler_exclusion.py
hexinw-nvidia: **Correction to this comment — please disregard the documentation request. The constant should be deleted, not documented.**

The freshness bound only changes behaviour in one of the three possible states:

| artifact state (possibly stale) | with the bound | without it |
| --- | --- | --- |
| missing | fail open -> join | fail open -> join |
| present, task not listed | join | join |
| present, task listed | stale -> fail open -> **join** | **leave** |

So its only effect is to stop honouring an *exclusion* after 30 minutes. That is only needed if an array index can be re-allocated to a different set of nodes. This integration already requires `--no-requeue` and an artifact directory keyed by `SLURM_ARRAY_JOB_ID`, so index K within array job J maps to exactly one allocation for the life of that job: a stale "excluded" entry cannot wrongly evict anything, because the task it names has already exited and nothing will ever run as that index again.

Note the bound was never protection against the producer dying — a dead producer leaves the artifact missing or stale-but-clean, and both of those fail open with or without it.

Suggest deleting `_DECISION_MAX_AGE_SECONDS`, the `modified_at` read in `_read_first_control_record`, and the `expired` outcome. What becomes load-bearing instead is the two preconditions, which should be stated in the integration guide as *requirements* rather than deployment advice. For a deployment that cannot hold them, stamping allocation identity into the artifact — a nodelist the consumer compares against `$SLURM_JOB_NODELIST`, failing open on mismatch — is an exact backstop where a time bound is only a probabilistic one.

namitdhameja:
@hexinw-nvidia I think, I got all of them - let me know if otherwise. Comments:
 - The initial implementation used `json.loads`. Strict regex parsing was explicitly requested, so I changed it accordingly. I obv preferred the JSON-based implementation as well, so I’ll revert to it.
 - The freshness bound of 30-min is valid. However, the practical risk is small and not worth prolonging the review. I’ve removed it.
 - nvrx_singleton_array.sbatch changes can be at the producer side

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
```diff
@@ -0,0 +1,145 @@
...
+
+@dataclass(frozen=True)
+class SlurmArrayTask:
+    """Identity needed to consume one Slurm array task's decision."""
+
+    job_id: str
+    task_id: str
+
+
+class SegmentHealthCheck:
```
hexinw-nvidia, line 19: **Proposed replacement for this module** (145 lines to ~55):

```python
"""Consume Scheduler Exclusion decisions before joining rendezvous.

Artifact: first line is the decision - the allocation IDs to exclude, whitespace
separated. Later lines are evidence for humans and are not read here.

    7 12
    7 node02
    12 node07,node09
"""

import logging
import os
from pathlib import Path
from typing import Optional

from nvidia_resiliency_ext.shared_utils.log_manager import LogConfig

log = logging.getLogger(LogConfig.name)

#: Sanity bound on the decision line. 313 ten-digit IDs is about 3.4 KiB.
_MAX_RECORD_BYTES = 8 * 1024


class SegmentHealthCheck:
    """Report this allocation unhealthy when the scheduler has excluded it.

    Fails open on anything unexpected: an exclusion we cannot read is never a reason
    to refuse a rendezvous. The caller (`_run_health_check`) logs the ERROR naming
    this check when it reports unhealthy, so nothing is logged on that path here.
    """

    def __init__(self, directory: str, job_id: str, task_id: str) -> None:
        self._path = Path(directory) / f"scheduler_exclusion.{job_id}"
        self._task_id = task_id

    def __call__(self) -> bool:
        return self._perform_health_check()

    def _perform_health_check(self) -> bool:
        try:
            # errors="replace": a corrupt artifact must not raise UnicodeDecodeError, which
            # is a ValueError and would escape the OSError handler below -- _run_health_check
            # would then turn it into UnhealthyNodeException and evict this node. Undecodable
            # input has to fail open like every other unreadable case.
            with self._path.open(encoding="utf-8", errors="replace") as stream:
                excluded = stream.readline(_MAX_RECORD_BYTES).split()
        except FileNotFoundError:
            return True  # producer has not published yet; fail open, quietly
        except OSError as exc:
            # Unreadable for a reason other than absence - permissions, EIO. Rare, and
            # at most one line per rendezvous, so worth surfacing.
            log.warning("Ignoring Scheduler Exclusion decision: %s", exc)
            return True
        return self._task_id not in excluded


def get_segment_health_check(directory: Optional[str]) -> Optional[SegmentHealthCheck]:
    """The check for this launcher, or None when it should not run.

    Only SLURM_PROCID=0 reads the artifact - one reader per array task, or per job for a
    non-array deployment.
    """
    if not directory or os.environ.get("SLURM_PROCID") != "0":
        return None

    job_id = os.environ.get("SLURM_ARRAY_JOB_ID") or os.environ.get("SLURM_JOB_ID")
    if not job_id:
        return None
    # Non-array deployment: the job itself is the unit of exclusion.
    task_id = os.environ.get("SLURM_ARRAY_TASK_ID") or job_id

    log.info("Scheduler Exclusion installed job_id=%s task_id=%s", job_id, task_id)
    return SegmentHealthCheck(directory, job_id, task_id)
```

What goes, and why:

- **`SlurmArrayTask` (24)** — two strings. Passing them directly also collapses the two-step construction in the barrier.
- **`_nonempty_env_value` (123)** — applied to 2 of the 5 env reads here; the other three are raw. Either all need whitespace normalisation or none do, and Slurm sets these cleanly.
- **`_read_first_control_record` (137)** — nine lines for one `readline`. Its "not newline terminated" branch is torn-write detection, and the producer publishes by atomic `mv`, so a reader can never observe a partial file.
- **`_decision_file_path` (130)** — the path is fixed for the object's life; build it once in `__init__`. This is also where the `.state` / `.jsonl` mismatch lives.
- **`_COMPACT_DECISION_PATTERN`, `_MAX_DECISION_LINE_BYTES`, the quoted byte token** — one `split()` plus a membership test. `split()` absorbs CRLF, tabs and any Slurm field padding, and field equality means `7` cannot match `17` by construction.
- **The `SLURM_NODEID`/`SLURM_PROCID` fallback (87-95)** — `srun` sets both together, so the `else` is unreachable. And because it *is* an `else`, under `NVRX_ENABLE_MULTI_LAUNCHERS_PER_NODE=1` every launcher on node 0 installs the check. `SLURM_PROCID == 0` picks exactly one launcher in both shapes, and is what the rest of NVRx keys node identity on — `SLURM_NODEID` appears nowhere else in the tree. (Not `get_infrastructure_rank()`: for arrays that is `array_task_id * nnodes + procid`, so rank 0 would select one launcher for the entire array rather than one per task.)
- **The per-check `log.info` with `elapsed_ms`** — `_run_health_check` already logs an ERROR naming this check on exclusion, and `perform_rendezvous` already times the whole health-check phase at debug and records `HEALTH_CHECK_COMPLETED`. This duplicated both, one level louder.
- **The `replacement_group_size` gate (105-121)** — this one is a judgement call rather than an obvious win, so push back if you disagree. The group ID *is* the array task ID (`_current_replacement_group_id` returns `SLURM_ARRAY_TASK_ID`), so one group per array task holds structurally; the guard asserts an invariant rather than establishing one, and when it trips it disables the feature on a single warning. If you want it kept, I would demote it to debug rather than a disable.

**What it adds:** non-array Slurm jobs. Today line 83 bails when `SLURM_ARRAY_TASK_ID` is absent, and the integration guide scopes regular jobs out in writing. `SLURM_ARRAY_TASK_ID or SLURM_JOB_ID` covers both shapes in one line — for a plain `sbatch` the job is the unit of exclusion. That does mean the producer needs to publish job IDs for non-array jobs; say if that is a problem on your side.

Happy to put this up as a PR against your branch if that is easier than transcribing it.

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -2500,6 +2508,26 @@ def __init__(
         self._storage_path_checker = (
             StoragePathHealthCheck(storage_healthcheck_paths) if storage_healthcheck_paths else None
         )
+        scheduler_exclusion_task = (
```
hexinw-nvidia: Two things in this file.

**This block becomes one line.** Twenty lines, two chained conditional expressions, and an intermediate that exists only to be passed from the first to the second and then read back for the log. With `job_id`/`task_id` passed directly:

```python
# None unless this launcher is the one reader for its allocation; see the factory.
self._segment_health_checker = get_segment_health_check(settings.scheduler_exclusion_dir)
```

The directory guard moves into the factory, the second conditional disappears with the intermediate, and the log moves to where `job_id`/`task_id` are already locals. It then reads like `_storage_path_checker` two lines above, and the null-check at the call site matches how `get_node_health_check()` is already handled — which is why the factory is named to match. Dropping the explicit `os.environ` argument also keeps env plumbing out of the constructor; tests inject with `patch.dict(os.environ, ...)`, which the barrier tests in this PR already do.

**Please revert the `SLURM_*` precedence change at line 3039.**

```python
-    nnodes_per_group = os.getenv("SLURM_NNODES", os.getenv("SLURM_JOB_NUM_NODES"))
+    nnodes_per_group = os.getenv("SLURM_JOB_NUM_NODES", os.getenv("SLURM_NNODES"))
```

`_get_replacement_group_size()` runs for **every** Slurm job array, whether or not scheduler exclusion is configured, and its result feeds active-rank assignment. Inside an srun step those two variables can disagree — which is exactly why I asked which one the *new* gate should prefer. As written this silently changes `replacement_group_size` derivation for existing users who are not using this feature at all.

If the reorder is right it deserves its own PR with its own justification and a test pinning the step-vs-job semantics. The docstring table re-padding at 3084-3090 is collateral from the longer name and goes with it — that is the entire `-7` in this diff.

hexinw-nvidia (changes requested):
Re-review after the rebase — base on `main`, squashed, `job_metadata.py` dropped, the third absolute-path check gone, the lease removed. Thanks, those all landed.

Two changes requested. I have other smaller notes but they can wait; these are the ones worth doing first.

**The feature is currently inert end to end.** This PR reads `scheduler_exclusion.<job_id>.state`; #399's producer writes `scheduler_exclusion.<job_id>.jsonl`. The consumer gets `FileNotFoundError` and fails open, on every task, always. Because it fails open, a QA run looks identical to a healthy run with no exclusions — worth fixing the naming in both PRs together.

1. `scheduler_exclusion.py` — proposed replacement below: 145 lines to ~55, and non-array Slurm jobs work, which they do not today.
2. `ft_rendezvous_barrier.py` — the install block collapses to one line, and the `SLURM_*` precedence change should come out of this PR.

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
namitdhameja: Order fixed now:
self._record(message=msg)
if self._segment_health_checker is not None:
    self._run_health_check(...)

src/nvidia_resiliency_ext/fault_tolerance/scheduler_exclusion.py
namitdhameja: moved to warning.

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
namitdhameja: Bounded-reads and _read_first_control_record: are defensive check, such as accidental file copy and thus a corrupt file. Its protection from events outside of the Producer. 

_decision_file_path: The prod and consumer use scheduler_exclusion.<job_id>.state, I didn't find a mismatch.

_COMPACT_DECISION_PATTERN: I missed the point of changing, so kept as-is.  

I think have got the rest. Pls flag if otherwise.

hexinw-nvidia:
Should we do the following renaming?

  scheduler_exclusion.py            -> segment_health_check.py                             
  --ft-scheduler-exclusion-dir      -> --ft-segment-health-check-dir                       
  scheduler_exclusion_dir           -> segment_health_check_dir     (config, RendezvousSettings, rdzv_configs key)
  integration/scheduler_exclusion.rst -> integration/segment_health_check.rst

namitdhameja:
> Member

I would say that are conflating between two:
1. Sch Exc: this is the feature
2. SegmentHC: That is how FT consumes said feature.

These 2 concepts are not at odds with each other; but I don't have a strong pref so can go with that.

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
```diff
@@ -0,0 +1,109 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Consume segment health decisions before joining rendezvous."""
+
+from __future__ import annotations
+
+import logging
+import os
+import re
```
hexinw-nvidia: Remove.

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
```diff
@@ -0,0 +1,109 @@
...
+from pathlib import Path
+from typing import Optional
+
+from nvidia_resiliency_ext.shared_utils.log_manager import LogConfig
+
+log = logging.getLogger(LogConfig.name)
+
+_MAX_DECISION_LINE_BYTES = 8 * 1024
+_COMPACT_DECISION_PATTERN = re.compile(rb'\[(?:"[0-9]+"(?:,"[0-9]+")*)?\]')
```
hexinw-nvidia: Remove

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
```diff
@@ -0,0 +1,109 @@
...
+        job_id,
+        task_id,
+    )
+    return SegmentHealthCheck(directory, job_id, task_id)
+
+
+def _decision_file_path(directory: str, job_id: str) -> Path:
+    base = Path(directory)
+    if Path(job_id).name != job_id:
+        raise ValueError(f"invalid Slurm job ID: {job_id!r}")
```
hexinw-nvidia: Nit: Remove. It is unnecessary over-defensive code.

hexinw-nvidia: The whole "_decision_file_path" can be removed. It can be called in the constructor only once. It doesn't need to be called in every health check.

You can do the following in the constructor.
self._path = Path(directory) / f"segment_health_check.{job_id}"

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
```diff
@@ -0,0 +1,109 @@
...
+        self.task_id = task_id
+
+    def __call__(self) -> bool:
+        return self._perform_health_check()
+
+    def _perform_health_check(self) -> bool:
+        """Evaluate the current segment health artifact, failing open."""
+        try:
+            decision_path = _decision_file_path(self.directory, self.job_id)
+            decision = _read_first_control_record(decision_path)
```
hexinw-nvidia: The whole logic in _read_first_control_record() is unnecessary over-defense. We can just do the following two lines:
            with self._path.open(encoding="utf-8", errors="replace") as stream:
                excluded = stream.readline(_MAX_RECORD_BYTES).split()

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
```diff
@@ -0,0 +1,109 @@
...
+            if _COMPACT_DECISION_PATTERN.fullmatch(decision) is None:
+                raise ValueError(
+                    "first segment health record must be a compact array of allocation IDs"
+                )
+
+            task_token = b'"' + self.task_id.encode("ascii") + b'"'
+            if task_token in decision:
+                return False
+            return True
```
hexinw-nvidia: Please remove.

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
```diff
@@ -0,0 +1,109 @@
...
+        except FileNotFoundError:
+            log.warning("Segment health decision is not available in %s", self.directory)
```
hexinw-nvidia: This will be too chatty when the producer doesn't run yet. We can inspect the lustre path without this message. Please remove and just do:
            return True  # producer has not published yet; fail open, quietly

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
```diff
@@ -0,0 +1,109 @@
...
+        except OSError as exc:
+            log.warning("Ignoring segment health decision: %s", exc)
```
hexinw-nvidia, line 38: Add the following after the log.warning.
           return True

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
```diff
@@ -0,0 +1,109 @@
...
+        except (TypeError, ValueError, UnicodeEncodeError) as exc:
+            log.warning("Ignoring segment health decision: %s", exc)
+        return True
```
hexinw-nvidia: Replace "return True" with the following
             return self._task_id not in excluded

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
```diff
@@ -0,0 +1,109 @@
...
+            task_token = b'"' + self.task_id.encode("ascii") + b'"'
+            if task_token in decision:
+                return False
+            return True
+        except FileNotFoundError:
+            log.warning("Segment health decision is not available in %s", self.directory)
+        except OSError as exc:
+            log.warning("Ignoring segment health decision: %s", exc)
+        except (TypeError, ValueError, UnicodeEncodeError) as exc:
+            log.warning("Ignoring segment health decision: %s", exc)
```
hexinw-nvidia: Please remove.

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
```diff
@@ -0,0 +1,109 @@
...
+        return True
+
+
+def get_segment_health_check(directory: Optional[str]) -> Optional[SegmentHealthCheck]:
+    """Return the segment health check for this launcher, when eligible."""
+    if not directory:
+        return None
+
+    if os.environ.get("SLURM_PROCID", "").strip() != "0":
+        return None
```
hexinw-nvidia: These can be simplified to the following. strip is not needed.
    if not directory or os.environ.get("SLURM_PROCID") != "0":
        return None

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
```diff
@@ -0,0 +1,109 @@
...
+    if array_job_id:
+        if not array_task_id:
+            log.warning(
+                "Segment health check is not installed because array-task process 0 is missing "
+                "SLURM_ARRAY_TASK_ID"
+            )
+            return None
+        task_id = array_task_id
+    else:
+        task_id = job_id
```
hexinw-nvidia: Unnecessary over-defense. Please simplify to:

    job_id = os.environ.get("SLURM_ARRAY_JOB_ID") or os.environ.get("SLURM_JOB_ID")
    if not job_id:
        return None
    # Non-array deployment: the job itself is the unit of exclusion.
    task_id = os.environ.get("SLURM_ARRAY_TASK_ID") or job_id

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
```diff
@@ -0,0 +1,109 @@
...
+
+def _read_first_control_record(path: Path) -> bytes:
+    with path.open("rb") as stream:
+        line = stream.readline(_MAX_DECISION_LINE_BYTES + 1)
+
+    if line.endswith(b"\n"):
+        return line[:-1]
+    if len(line) > _MAX_DECISION_LINE_BYTES:
+        raise ValueError(f"first decision record exceeds {_MAX_DECISION_LINE_BYTES} bytes: {path}")
+    raise ValueError(f"first decision record is not newline terminated: {path}")
```
hexinw-nvidia: Please remove

hexinw-nvidia (changes requested):

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
namitdhameja: I think it would be more helpful if are descriptive of what are trying to achieve here.

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
namitdhameja: I think it would be more helpful if are descriptive of what are trying to achieve here.

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
namitdhameja: Perhaps you can walk thru the concern here.

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
hexinw-nvidia: We will be comparing the value against the TaskID. If the value inside the first line is incorrect (non-numeric) for whatever reason, InJob will safely ignore and proceed.

The whole RE check is unnecessarily over-defensive. In fact, it will raise a ValueError and crash InJob.

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
hexinw-nvidia: self.task_id is derived from SLURM env as a string. It is expected to be the same type of value as reported from the producer. This goes back to my previous question about the file format.

I'd expect the first line format to be a space-delimited list of task IDs. No need to have double quote "" around the number.

The whole line 43 should not be done in every health check. It should be constructed in the constructor if needed. But overall, the whole line 43 is not needed once we agreed the first line to be just a space-delimited list of task IDs (no quotes).

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
hexinw-nvidia: Why do we need to enforce the line ends with "\n" in line 81?
Line 83-84 are dead code, given we already do a cap-limit in the read. Correct? Am I missing something?

namitdhameja: "\n" here means that a complete record is read, for protection against external corruption. Eg, a random executable is copied over. We will read 8k, but determine it to be a truncated record and not further process it. Its a simple defensive check against partial or malformed data.

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
namitdhameja: The first record should be a grepable list of task IDs, as discussed in Slack, and the current format satisfies that requirement. We did not specify whether the delimiter should be spaces, commas, colons, or something else, because the delimiter does not materially affect functionality. I am therefore unclear on the motivation for changing from commas to spaces. In my view, the current format is faithful to what we agreed.

hexinw-nvidia: A plain text like the following:
`1 42 100`

is simpler and more natural than:
`"1" "42" "100"`

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
namitdhameja: A non-numeric value fails the regex and raises `ValueError`, but that exception is caught. The check logs a warning and returns `True`, so InJob proceeds and does not crash. I am also unclear about “unnecessarily over-defensive”; any file-based scheme must safely handle malformed or corrupted input.

src/nvidia_resiliency_ext/fault_tolerance/segment_health_check.py
namitdhameja: Comma-separated values are conventional and compact. If the concern is specifically the quotes, then those will remove.

hexinw-nvidia: Agreed. Quotes are not needed. Comma-separated values are fine, though there is a nuance between comma "," and space " " when coming to Python split method. The comments and replies are becoming too lengthy. I will start a fresh message to address your response.

hexinw-nvidia:
Starting a new thread to summarize what I meant by a simpler implementation. 
**Design assumption:**

1. The first line of the report file produced by the producer is a comma-delimited list of task_id values. Each task_id is a numeric integer value provided by SLURM.
2. The producer always updates the report file atomically. The consumer should not need to handle cases where it reads the file while the producer is in the middle of writing it. Conceptually, treat the file as a key/value object in the data store, where updating the value for the key is always atomic.

The current block of code (let's name it **Code1**) is functionally correct.

    _COMPACT_DECISION_PATTERN = re.compile(rb'\[(?:"[0-9]+"(?:,"[0-9]+")*)?\]')

    # some class method
    def _perform_health_check(self) -> bool:
        """Evaluate the current segment health artifact, failing open."""
        try:
            decision = _read_first_control_record(self._path)
            if _COMPACT_DECISION_PATTERN.fullmatch(decision) is None:
                raise ValueError(
                    "first segment health record must be a compact array of allocation IDs"
                )

            task_token = b'"' + self.task_id.encode("ascii") + b'"'
            if task_token in decision:
                return False
            return True
        except FileNotFoundError:
            return True
        except OSError as exc:
            log.warning("Ignoring segment health decision: %s", exc)
            return True
        except (TypeError, ValueError, UnicodeEncodeError) as exc:
            log.warning("Ignoring segment health decision: %s", exc)
        return True

    # A helper method
    def _read_first_control_record(path: Path) -> bytes:
        with path.open("rb") as stream:
            line = stream.readline(_MAX_DECISION_LINE_BYTES + 1)
        if line.endswith(b"\n"):
            return line[:-1]
        if len(line) > _MAX_DECISION_LINE_BYTES:
            raise ValueError(f"first decision record exceeds {_MAX_DECISION_LINE_BYTES} bytes: {path}")
        raise ValueError(f"first decision record is not newline terminated: {path}")`

**Code1** can be safely replaced by the following code block **Code2**

    def _perform_health_check(self) -> bool:
        try:
            # errors="replace": a corrupt artifact must not raise UnicodeDecodeError, which
            # is a ValueError and would escape the OSError handler below -- _run_health_check
            # would then turn it into UnhealthyNodeException and evict this node. Undecodable
            # input has to fail open like every other unreadable case.
            with self._path.open(encoding="utf-8", errors="replace") as stream:
                excluded = stream.readline(_MAX_RECORD_BYTES).rsplit("\n").split(",")
        except FileNotFoundError:
            return True  # producer has not published yet; fail open, quietly
        except OSError as exc:
            # Unreadable for a reason other than absence - permissions, EIO. Rare, and
            # at most one line per rendezvous, so worth surfacing.
            log.warning("Ignoring Scheduler Exclusion decision: %s", exc)
            return True
        return self._task_id not in excluded

**Code2** IMO is simpler and cleaner:

1. No format validation is required. The split method splits arbitrary content. Garbage is treated as an exclusion list.
2. No newline termination check. There should not be a producer mid-write.
3. No over-length guard is needed. The "readline(_MAX_RECORD_BYTES)" will safely truncate if the content ever becomes so long. 
4. No helper method is needed.

Now come to the comment of space vs comma as the delimiter. For a space-delimited list of task_ids, we can just do:
    excluded = stream.readline(_MAX_RECORD_BYTES).split()
    
But as I said, I am not hung up on the comma vs space.

hexinw-nvidia (approved):
LG on the per-task files implementation. 
Thanks for making the changes.

PR #399, opened 2026-08-09
Add Bash Scheduler Segment Health producer

namitdhameja:
## Summary

- add one sourceable Bash Scheduler Segment Health producer for Slurm array task 0
- query scheduler-unavailable nodes with partition-filtered `sinfo`
- skip `squeue` entirely when there are no unavailable nodes and no prior decisions
- otherwise query all running array tasks, expand each compressed hostlist locally, and map unavailable nodes to exact task IDs
- publish per-task CSV control markers, `.inactive` last-observation artifacts, and centralized best-effort JSONL history
- preserve prior decisions after incomplete scheduler results or publication failures; clear stale decisions after the configured freshness window
- integrate the producer lifecycle into the singleton-array deployment example without creating a daemon or supervising the workload

## Motivation

The producer runs in the host-side sbatch control flow where Slurm commands are available. The NVRx workload can remain inside its container and consume the resulting per-task files from shared storage through `SegmentHealthCheck`.

The producer is one sourceable Bash file with three operations: `scheduler_segment_health_configure`, `scheduler_segment_health_poll_once`, and `scheduler_segment_health_cleanup`. The surrounding sbatch control loop owns cadence and lifecycle.

## Algorithm

1. A partition-filtered `sinfo` query produces the canonical unavailable-node set for `DRAIN`, `DOWN`, `FAIL`, and `NO_RESPOND` states.
2. If that set and prior published state are both empty, the poll completes without calling `squeue`.
3. Otherwise, one array-wide `squeue` returns `%K|%N` for every running task.
4. Each compressed hostlist is expanded locally with `scontrol show hostnames`, normalized, and intersected with the unavailable-node set.
5. Only after a complete result is derived does reconciliation set, update, clear, or retire individual task artifacts.
6. A tracked task absent from two consecutive complete `squeue` results moves from the current namespace to `.inactive`; one missing result preserves its current decision.

The reconciliation path scales with the number of running array tasks and their total allocated nodes. For `R` running tasks, `U` unavailable nodes, and `N_i` nodes in task `i`, its local set-processing work is approximately `O(U log U + sum_i(N_i log N_i + U))`. It is not independent of total allocated nodes. The healthy fast path performs only `sinfo`.

## Contract

```text
segment_health_check.<array_job_id>.<task_id>
segment_health_check.<array_job_id>.<task_id>.inactive
segment_health_check_history.<array_job_id>.log
```

A non-empty current task file excludes that task. Its sorted unavailable-node CSV is diagnostic context; the consumer checks only that the file is a non-empty regular file. A missing or zero-byte current file means no current exclusion. The centralized log records `excluded`, `updated`, and `cleared` transitions as JSONL.

The deployment uses `--no-requeue` and scopes the artifact directory to `SLURM_ARRAY_JOB_ID`.

## Validation

- local producer unit suite: 33 passed
- exact pushed commit `a73e2bb` on Linux/PDX: 33 passed, Slurm job `3551`
- dev FT baseline: passed, job `3552`
- exclusion without spare: passed, job `3554`
- rendezvous-host exclusion: passed, job `3556`
- recovery and marker clear: passed, job `3558`
- complete-spare replacement: passed, job `3560`
- state reconstruction, update, clear, and two-poll inactive transition: passed, job `3563`
- all mutated nodes restored to `idle`; no qualification jobs remain
- `bash -n`, `black --check`, `isort --check-only`, and `git diff --check` passed

### DIFF
--- a/examples/fault_tolerance/deployment/slurm/README.md
+++ b/examples/fault_tolerance/deployment/slurm/README.md
@@ -7,6 +7,7 @@ training rank and the rendezvous host. See [../README.md](../README.md) for why.
 ```
 nvrx_singleton_array.sbatch   one generation: rendezvous, gate, ft_launcher, teardown
 submit_chain.sh               computes the array shape and enqueues K generations
+scheduler_segment_health.sh    optional: publishes scheduler-unavailable array tasks
 ```
 
 ## Requirements
@@ -203,3 +204,88 @@ neither the job nor `squeue` will tell you about.
    precedes the exit that triggers the teardown, and any node that exited cleanly is a
    better witness than task 0's own code. The flag is per-generation and needs no cleanup;
    `cancel_chain` remains the one mechanism that stops the successors.
+---
+
+## Scheduler segment-health producer
+
+`scheduler_segment_health.sh` exposes a sourceable interface:
+`scheduler_segment_health_configure` validates and initializes configuration
+without querying Slurm, `scheduler_segment_health_poll_once` performs one poll,
+and `scheduler_segment_health_cleanup` releases producer-owned temporary state
+at caller exit. The producer publishes non-empty per-task control markers
+consumed by `SegmentHealthCheck`. The caller owns cadence and handles a failed
+pass:
+
+```bash
+if [[ "${SLURM_ARRAY_TASK_ID}" == "0" ]]; then
+  export NVRX_SEGMENT_HEALTH_CHECK_DIR="<shared-dir>/${SLURM_ARRAY_JOB_ID}/scheduler-segment-health"
+  export SLURM_JOB_PARTITION
+  source scheduler_segment_health.sh
+
+  scheduler_segment_health_configure || exit 1
+  scheduler_segment_health_poll_once || \
+    echo "scheduler segment-health poll failed; prior decision preserved" >&2
+fi
+```
+
+`nvrx_singleton_array.sbatch` integrates this lifecycle directly. Setting
+`NVRX_SEGMENT_HEALTH_CHECK_DIR` enables the feature; leaving it empty preserves
+the existing launch path. `submit_chain.sh` copies the producer into the run
+directory, and the example's host-side control loop invokes
+`scheduler_segment_health_poll_once` and performs best-effort cleanup when that
+loop exits. The producer itself is a one-shot function: it does not create a
+loop or background process. Every task receives the same
+`--ft-segment-health-check-dir` value:
+
+```bash
+NVRX_SEGMENT_HEALTH_CHECK_DIR=/lustre/run/segment-health ./submit_chain.sh
+```
+
+The example control-loop interval is 600 seconds and can be changed with
+`NVRX_CONTROL_LOOP_INTERVAL_SECONDS`. A direct `sbatch` submission must
+place `scheduler_segment_health.sh` under `NVRX_WORK_DIR` or set
+`NVRX_SEGMENT_HEALTH_CHECK_PRODUCER` to its readable path.
+
+### Poll and publication behavior
+
+Every poll performs one partition-filtered `sinfo`. When unavailable nodes or
+existing active decisions exist, one array-wide `squeue` builds the complete
+running task-to-node map. Local `scontrol show hostnames` expansion does not
+query `slurmctld` for ordinary bracket hostlists.
+
+The producer maps nodes in `DRAIN`, `DOWN`, `FAIL`, or `NO_RESPOND` state and
+publishes:
+
+```text
+segment_health_check.<array_job_id>.<task_id>
+segment_health_check.<array_job_id>.<task_id>.inactive
+segment_health_check_history.<array_job_id>.log
+```
+
+A non-empty current task file excludes its task; zero bytes or absence means no
+current exclusion. Its canonical unavailable-node CSV is diagnostic context.
+When a tracked task is absent from two consecutive complete `squeue` results,
+its unsuffixed file is renamed to `.inactive`. One omitted result preserves the
+current decision, guarding against a transiently incomplete scheduler response.
+The producer keeps a process-local set of tasks missing from the prior complete
+poll. Current absence is derived from the new running-task map: a task already
+in the prior set moves inactive, while a first miss becomes the set used by the
+next poll. A producer restart resets this one-poll memory.
+The centralized `.log` is best-effort JSONL audit history.
+
+Each poll reconstructs prior task state from the unsuffixed task files, then changes
+them only after deriving a complete scheduler result. Failed polls preserve
+prior decisions until 30 minutes have elapsed since the last complete poll;
+the producer then clears non-empty current task files and fails open. Per-task
+publication errors do not block other tasks and are reconciled by the next
+invocation's filesystem scan. `sinfo` and `squeue` have 30-second timeouts. The
+surrounding `sbatch` control flow owns invocation cadence.
+
+The clean path performs only `sinfo`: when there are no unavailable nodes and
+no prior decisions, it skips `squeue` and hostlist expansion. Reconciliation
+queries all running array tasks and expands their hostlists locally, so that
+path scales with the number of running tasks and their total allocated nodes.
+
+The initial deployment requires an explicitly non-requeued array and an output
+directory scoped to `SLURM_ARRAY_JOB_ID`. Producer logs remain in batch
+stdout/stderr.
--- a/examples/fault_tolerance/deployment/slurm/nvrx_singleton_array.sbatch
+++ b/examples/fault_tolerance/deployment/slurm/nvrx_singleton_array.sbatch
@@ -69,6 +69,11 @@ NVRX_MODEL_PROFILE="${NVRX_MODEL_PROFILE:-small}"
 
 # Optional. Endpoint of the node health check daemon; omitted when empty.
 NVRX_HEALTH_CHECK_ENDPOINT="${NVRX_HEALTH_CHECK_ENDPOINT:-}"
+# Optional. A shared artifact directory enables Scheduler Segment Health for
+# both the task-0 producer and every ft_launcher consumer. Empty disables it.
+NVRX_SEGMENT_HEALTH_CHECK_DIR="${NVRX_SEGMENT_HEALTH_CHECK_DIR:-}"
+NVRX_CONTROL_LOOP_INTERVAL_SECONDS="${NVRX_CONTROL_LOOP_INTERVAL_SECONDS:-600}"
+NVRX_SEGMENT_HEALTH_CHECK_PRODUCER="${NVRX_SEGMENT_HEALTH_CHECK_PRODUCER:-${NVRX_WORK_DIR}/scheduler_segment_health.sh}"
 # Minimum nodes per NVLink domain, identified by GPU ClusterUUID -- a restart cycle then
 # only ever selects complete segments. 1 suits GB200 NVL72 and other NVLink-domain
 # systems (DGX H200, HGX B200). Set empty on systems with no ClusterUUID (e.g. H100),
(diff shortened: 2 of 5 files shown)


examples/fault_tolerance/deployment/slurm/scheduler_exclusion_poller.sh
```diff
@@ -0,0 +1,260 @@
...
+        "${ARRAY_JOB_ID}" "${observed_at}" >>"${destination}"
+    while IFS= read -r task_id; do
+        [[ -n "${task_id}" ]] || continue
+        printf '%s{"task_id":"%s","restart_count":0,"valid_until":"%s"}' \
+            "${separator}" "${task_id}" "${valid_until}" >>"${destination}"
+        separator=,
+    done <"${tasks}"
+    printf ']}\n' >>"${destination}"
+
+    printf '{"type":"decision","schema_version":1,"job_id":"%s","generated_at":"%s","scope":"node","excluded_nodes":[]}\n' \
```
hexinw-nvidia, line 129: **Blocker.**

`excluded_nodes` is hardcoded empty, so nothing downstream can answer "which node caused task 7 to be excluded?" — not an operator reading the artifact, not the consumer's log line. This is the piece I need before merge.

The producer already holds both sets at this point. `-o '%K|%N'` on the squeue at line 177, then the intersection `drain_poller.sh` already does (`scontrol show hostnames "${nodelist}" | grep -Fxf "${unavailable}"`) — about eight lines. Note `scontrol show hostnames` is a *local* expansion, no controller RPC.

Two extra failure modes come with it and need the same preserve-and-retry treatment as the rest of the pass: `scontrol` failing, and an empty intersection for a task squeue said was affected.

examples/fault_tolerance/deployment/slurm/scheduler_exclusion_poller.sh
```diff
@@ -0,0 +1,260 @@
...
+    printf '[' >>"${destination}"
+    while IFS= read -r task_id; do
+        [[ -n "${task_id}" ]] || continue
+        printf '%s"%s"' "${separator}" "${task_id}" >>"${destination}"
+        separator=,
+    done <"${tasks}"
+    printf ']\n' >>"${destination}"
+}
+
+write_artifact() {
```
hexinw-nvidia, line 108: **Blocker** (applies to lines 58, 76, 85, 111, 121 as well).

With #358 deprecated, records 2 and 3 have no reader. I grepped the tree: the artifact's only consumer is `SegmentHealthCheck`, which calls `readline()` once and never touches anything after line 1.

Please drop them, along with `format_timestamp`, `DECISION_TTL_SECONDS`, `valid_until`, and the `observed_epoch`/`valid_epoch` derivation at line 201 with its error branches. `valid_until` is worse than unused — freshness is actually enforced by file mtime on the consumer side, so the field advertises a lease that nothing honours.

Combined with the blocker above, the artifact becomes one line carrying both the decision and its evidence:

```json
{"7":["node02"],"12":["node07","node09"]}
```

The consumer needs no coordinated change for this: `"7" in decision` behaves identically whether `decision` is the current list or this object, so #398 can merge on the array shape and #399 can switch whenever.

examples/fault_tolerance/deployment/slurm/scheduler_exclusion_poller.sh
```diff
@@ -0,0 +1,260 @@
...
+        || ! valid_until=$(format_timestamp "${valid_epoch}"); then
+        log "could not format artifact timestamps; preserving the previous decision"
+        return 1
+    fi
+    if ! write_artifact "${excluded}" "${observed_at}" "${valid_until}" \
+        "${decision}"; then
+        log "could not stage the decision artifact; preserving the previous decision"
+        return 1
+    fi
+    if ! head -n 1 "${decision}" >"${current_first}" \
```
hexinw-nvidia, line 227: Keep this behaviour, simplify the mechanism.

Once the file is a single line, `wc -c` on the file replaces the `head -n 1` / `current_first` / `previous_first` / `cmp` / `changed` machinery, and the `FIRST`-gated "refreshed" log can go with it.

Important: **keep the unconditional republish.** The consumer's 30-minute freshness bound is measured from this file's mtime, so a `publish_if_changed` optimisation — exactly what `drain_poller.sh` does — would silently disarm every consumer 30 minutes after the cluster goes quiet, with healthy-looking producer logs. Worth stating that as a contract in the header comment so nobody "optimises" it later.

examples/fault_tolerance/deployment/slurm/scheduler_exclusion_poller.sh
```diff
@@ -0,0 +1,260 @@
...
+
+    if ! timeout "${SLURM_CMD_TIMEOUT}" \
+        sinfo --states="${EXCLUSION_STATES}" --Node --noheader --format='%N' \
+        >"${unavailable_raw}"; then
+        log "sinfo failed or timed out; preserving the previous decision"
+        return 1
+    fi
+
+    : >"${unavailable_staged}"
+    while IFS= read -r node; do
```
hexinw-nvidia, line 165: **Medium** (same at line 187).

Both loops use `IFS=` / `IFS='|'`, which preserve whitespace, and then hard-reject on regex mismatch and abort the pass. If any Slurm build pads `%N` or `%K` to a default field width, this producer never publishes — not once, ever — and every consumer silently fails open 30 minutes later. That is a permanent-stale failure, not a fail-open one, and the test doubles cannot catch it.

Cheap insurance: drop the `IFS=` on this loop so the default IFS trims, and `task_id="${task_id//[[:space:]]/}"` before the match at line 187. Also worth confirming `%K` padding against a real controller — I could not from here.

examples/fault_tolerance/deployment/slurm/scheduler_exclusion_poller.sh
```diff
@@ -0,0 +1,260 @@
...
+        return 1
+    fi
+
+    if [[ -s "${unavailable}" ]]; then
+        if ! unavailable_filter=$(paste -s -d ',' "${unavailable}"); then
+            log "could not build the unavailable-node filter; preserving the previous decision"
+            return 1
+        fi
+        if ! taskmap=$(timeout "${SLURM_CMD_TIMEOUT}" \
+            squeue -j "${ARRAY_JOB_ID}" -r -h -t R \
```
hexinw-nvidia, line 186: **Medium.**

`--nodelist` gets every drained node in the *cluster*, including other partitions, joined into one argument every pass. Beyond controller load, it couples this job's artifact to cluster-wide state: one node name that `sinfo` reports but `squeue -w` rejects — removed from config between the two calls, or not visible to this user — fails the whole pass.

`drain_poller.sh` avoids this by pulling the job's own task->nodelist map and intersecting locally, bounded by the job. Since the blocker above needs `%N` anyway, the local intersection comes almost for free.

examples/fault_tolerance/deployment/slurm/scheduler_exclusion_poller.sh
```diff
@@ -0,0 +1,260 @@
...
+
+if [[ ! "${ARRAY_JOB_ID}" =~ ^[0-9]+$ ]]; then
+    printf 'scheduler_exclusion_poller: ARRAY_JOB_ID must be numeric: %s\n' \
+        "${ARRAY_JOB_ID}" >&2
+    exit 2
+fi
+
+POLL_INTERVAL="${POLL_INTERVAL:-600}"
+EXCLUSION_STATES="${EXCLUSION_STATES:-drain,down,fail,no_respond}"
+SLURM_CMD_TIMEOUT="${SLURM_CMD_TIMEOUT:-30}"
```
hexinw-nvidia, line 56: Minor (with line 69).

`SLURM_CMD_TIMEOUT` and `EXCLUSION_STATES` are unvalidated while `POLL_INTERVAL` and `ARRAY_JOB_ID` are; a bad value there fails every pass silently.

And `mktemp -d` inside `OUTPUT_DIR` (line 69) leaks `.scheduler_exclusion.<id>.XXXXXX` scratch dirs into the shared directory on SIGKILL — `drain_poller.sh` uses a fixed `.work` for this reason.

hexinw-nvidia (commented):
Thanks for splitting producer and consumer, and for moving this off the Python service to plain bash — that was the right call, and the error handling (all-or-nothing passes, preserve-on-failure) is solid.

Two blockers below, and they compose: fixing the first makes the second easier rather than harder.

Now that #358 is deprecated there is no second producer to stay format-compatible with, and nothing anywhere reads records 2 and 3. I trimmed the script locally to confirm the size of the win before asking for it: **194 -> 89 code lines (-54%), with drained-node reporting added, not omitted.** More disappears than just the writers — the `head -n 1` / `current_first` / `previous_first` / `cmp` / `changed` / `FIRST` apparatus exists only to compare the first line of a *multi*-line file, and goes away with it.

Separately: I still think the per-task-file shape (as in `drain_poller.sh`) is the better contract, but I am not asking to reopen that here. I will file it as a follow-up design discussion for iteration 2. Everything below is within the current single-file design.

examples/fault_tolerance/deployment/slurm/scheduler_exclusion_poller.sh
hexinw-nvidia: **Correction to this comment — the constraint I described no longer applies.**

I asked you to keep the unconditional republish because the consumer's 30-minute bound is measured from this file's mtime. That bound is being dropped (see the companion comment on #398): it only ever affected the case of a stale *exclusion*, which cannot wrongly evict under the `--no-requeue` plus job-scoped-directory preconditions this integration already requires.

With no mtime lease, this producer is free to skip the write when the decision is unchanged, the way `drain_poller.sh` does with `publish_if_changed` — which is the better behaviour on Lustre. The `changed` tracking already here can drive that skip rather than only the logging.

The rest of the original comment stands: with a single-line file, `wc -c` on the file replaces the `head -n 1` / `current_first` / `previous_first` comparison.

Apologies for the churn — better to correct it now than leave you implementing a constraint we are removing.

hexinw-nvidia:
## Proposal: drop JSON from the artifact entirely

Following on from the two blockers and the freshness-lease correction above — with the audit records gone, the first line is the whole contract, and it does not need to be JSON. Plain text, one line per excluded task, `<task_id> <drained_nodes>`:

```
7 node02
12 node07,node09
```

Producer, replacing `write_artifact` / `write_compact_decision` / the `sed 's/[^,]*/"&"/g'` quoting:

```bash
printf '%s %s\n' "${task_id}" "${drained}" >>"${decision}"
```

Consumer (#398), replacing the regex, the quoted byte token and the bounded readline:

```python
for line in path.read_text().splitlines():
    f = line.split()
    if f and f[0] == task_id:
        return False          # excluded; f[1:] is the evidence, for the log line
return True
```

**Why this is sound and not just shorter:** this script already validates node names against `^[A-Za-z0-9._-]+$` and task IDs against `^[0-9]+$`, and aborts the pass on anything else. Neither domain admits whitespace, so whitespace is a safe delimiter with no escaping. JSON's role in this contract is escaping a problem the validation already makes impossible.

Measured against the encodings that silently fail open under the current byte regex:

```
canonical            task7=EXCL  evidence=node02   task17=ok  task9=ok
CRLF                 task7=EXCL  evidence=node02   task17=ok  task9=ok
extra whitespace     task7=EXCL  evidence=node02   task17=ok  task9=ok
tabs                 task7=EXCL  evidence=node02   task17=ok  task9=ok
no trailing newline  task7=EXCL  evidence=node02   task17=ok  task9=ok
empty file           task7=ok    evidence=-        task17=ok  task9=ok
truncated mid-line   task7=EXCL  evidence=node02   task17=ok  task9=ok
```

All of them pass. `str.split()` absorbs `\r`, tabs and runs of spaces, so CRLF and field padding are handled for free — **which also retires my earlier comment about Slurm `%K`/`%N` padding on lines 157/187**: a padded field now parses correctly instead of failing the pass. `7` vs `17` is exact field equality rather than substring search, so that reasoning goes too. A crash-truncated final line costs one task, not the whole file.

What this removes across both PRs:

- here: `format_timestamp`, `write_compact_decision`, `write_artifact`, the JSON quoting, `MAX_DECISION_LINE_BYTES`, the `head -n 1` / `current_first` / `previous_first` comparison;
- in #398: `_COMPACT_DECISION_PATTERN`, `_MAX_DECISION_LINE_BYTES`, `_read_first_control_record`'s bounded readline and its two `ValueError` paths, the `task_token` construction, and `UnicodeEncodeError` from the except clause.

Two follow-ups if you take it: rename the artifact, since `.jsonl` becomes inaccurate (`scheduler_exclusion.<array_job_id>`, no extension, or `.txt`); and the evidence field satisfies the drained-node blocker in the same line as the decision, with no nested structure.

Happy to open this as a PR against your branch if that is easier than folding it in.

hexinw-nvidia (approved):

PR #401, opened 2026-08-10
Lens telemetry onto main

rhewett-nv:


src/nvidia_resiliency_ext/checkpointing/async_ckpt/core.py
```diff
@@ -750,6 +764,71 @@ def _handle_sigterm(signum, frame):
...
+            # the worker. Absent (older caller), the base groups already set stand.
+            _resolved = otel_bootstrap.get('resolved_span_groups')
+            if _resolved:
+                try:
+                    from nemo.lens.state import set_enabled_span_groups
+
+                    set_enabled_span_groups(frozenset(_resolved))
+                except ImportError:
+                    pass
+
```
shurkat-nvidia, line 803: This shouldn't have been embedded as a mandatory export format, I think

src/nvidia_resiliency_ext/fault_tolerance/launcher.py
```diff
@@ -1107,6 +1275,20 @@ def _start_workers(self, worker_group: WorkerGroup) -> Dict[int, Any]:
 
         current_cycle_info_path = self._current_cycle_info_path()
 
+        # One launch stamp for this worker cohort, so every restart hands its workers a fresh
+        # per-restart anchor. The batch script's launch_script_start is set once, outside the
+        # single srun, and is stale for restarts; NVRX_LAUNCH_TIME is the in-srun equivalent each
+        # cohort actually needs. The trainer uses it as its per-restart launch anchor, so its
+        # startup span measures this relaunch (rendezvous rejoin, spawn, import) inside this
+        # cycle's run span.
+        _nvrx_cohort_launch = repr(time.time())
```
shurkat-nvidia, line 1284: `repr` seems fragile? Shouldn't we use `strftime` or something similar to format this properly?

src/nvidia_resiliency_ext/shared_utils/_otel_fallbacks.py
```diff
@@ -0,0 +1,74 @@
...
+
+            return metrics.get_meter(__name__)
+
+        def shutdown(self, timeout_ms=5000):
+            pass
+
+    @contextmanager
+    def managed_span(group, name, tracer=None, **attributes):
+        """No-op context manager; yields None."""
+        yield None
```
shurkat-nvidia, line 74: Not a huge fan of redefining Nemo APIs here. I think the right approach may be to define our own spans that optionally call into Nemo when Nemo is present, nop otherwise.

rhewett-nv: I agree -- But I think that is a change that can be done in the background in the future.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/core.py
rhewett-nv: removed

hexinw-nvidia:
@shurkat-nvidia 's review comments will be incorporated into new PR#402.

PR #402, opened 2026-08-10
Lens telemetry, restructured into shared_utils/telemetry

hexinw-nvidia:
Takes over #401 (Russell is PTO) and restructures the OTel/nemo-lens integration along ownership lines. #401's five commits are included unmodified; the last commit is the refactor.

## Why

In #401 the telemetry logic lived inline in the code it instruments: ~330 lines of span-tree state machine on `FaultToleranceProfiler`, ~200 lines of setup/annotate/stage/shutdown in `launcher.py`, and a 60-line bootstrap block in the checkpoint worker's process target. Each call site carried its own `if enabled` guard and `try/except`.

## What changed

New `src/nvidia_resiliency_ext/shared_utils/telemetry/` package, layered innermost first:

| Module | Owns |
|---|---|
| `backend.py` | The only module that knows nemo-lens import paths (was `shared_utils/_otel_fallbacks.py`) |
| `spans.py` | `CycleSpanTree` — *how* spans nest, open and close. Best-effort by construction |
| `recorder.py` | `RestartCycleRecorder` — *when*, as a declarative `ProfilingEvent -> CycleAction` table |
| `agent.py` | `LauncherTelemetry` — the facade the ft_launcher agent codes against; a null object when telemetry is off |
| `checkpoint_worker.py` | `CheckpointWorkerTelemetry` — the spawned checkpoint worker's facade |

The key inversion: `profiling.py` gains a generic observer hook (`ProfilingListener`, `ProfilingEventRecord`, `add`/`remove_profiling_listener`) and a `profiling_phase` context manager, and no longer knows OTel exists. The recorder *observes* the profiler rather than being driven by the launcher.

Effect on the instrumented files:

- `launcher.py`: +201 -> +15 lines. Inline `setup_telemetry`, four `_otel_*` helpers, the shutdown thread and the worker-env block collapse to `LauncherTelemetry.create(...)` plus five one-line calls. The null object removes every guard and per-call-site `try/except`.
- `profiling.py`: +331 -> +62 lines, all of it generic.
- `ft_rendezvous_barrier.py`: the explicit `try/finally` around the Step-0 wait becomes `with profiling_phase(AWAIT_ROUND_STARTED, AWAIT_ROUND_COMPLETED, ...)`.
- `checkpointing/async_ckpt/core.py`: the inline bootstrap moves behind `CheckpointWorkerTelemetry.from_bootstrap()`.

Behavior is preserved: same span names, attributes, sweep semantics, flush points, cold-start backdating and worker env vars. Two deliberate small changes: checkpoint-worker telemetry setup failure no longer propagates (it now degrades the way the launcher's already did), and the agent deregisters its listener on shutdown.

## Tests and docs

- New `tests/shared_utils/test_telemetry_recorder.py` — 35 tests covering the span-tree state machine (full cycle, phase sweep, evicted node, standby/await, staged outcomes, cold start, flush points, profiler wiring). Runs without opentelemetry or nemo-lens installed: the recorder only needs an object with `start_span()`.
- New `shared_utils/telemetry/README.md` documenting the layering, the emitted span schema and how to add an event.

## Verification

- `black --check`, `isort --check-only`, `ruff check`: clean. (Note `launcher.py` carries `# fmt: off`, so black never saw #401's misformatted lines there; rewritten regardless.)
- `tests/fault_tolerance/unit/`: 542 passed. `tests/shared_utils/`: 154 passed plus the 35 new.
- Remaining local failures are environmental and identical before the change: 8 need the `ft_launcher` console script installed, 5 need NVML.
- `tests/checkpointing/unit/` needs GPUs and could not be run locally; `core.py` imports clean but that path is unverified by execution.

## Note

`shared_utils._otel_fallbacks` is gone, moved to `shared_utils.telemetry.backend`. It was private and never released, so no shim was left.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,6 +60,8 @@ pydantic-settings = ">=2.0.0"
 slowapi = ">=0.1.9"
 slack-bolt = { version = ">=1.23.0", optional = true }
 slack-sdk = { version = ">=3.35.0", optional = true }
+# Optional telemetry backend, pulled in by the `otel` extra.
+nemo-lens = { version = ">=0.2.0", extras = ["sdk"], optional = true }
 grpcio = "^1.76.0"
 grpcio-tools = "^1.76.0"
 protobuf = ">=4.22.0"
@@ -73,6 +75,9 @@ attribution = [
     "setproctitle",
     "slack-bolt",
     "slack-sdk",
+]
+otel = [
+    "nemo-lens",
 ]
 
 [tool.poetry.scripts]
--- a/src/nvidia_resiliency_ext/checkpointing/async_ckpt/core.py
+++ b/src/nvidia_resiliency_ext/checkpointing/async_ckpt/core.py
@@ -34,6 +34,7 @@ import torch
 from torch import multiprocessing as mp
 
 from ..utils import _disable_gc, debug_time
+from .telemetry import CheckpointWorkerTelemetry
 
 logger = logging.getLogger(__name__)
 
@@ -454,6 +455,7 @@ class PersistentAsyncCaller(AsyncCaller):
         io_priority: Optional[int] = None,
         sigterm_timeout: float = 30.0,
         cpu_shm_mode: bool = False,
+        otel_bootstrap: Optional[dict] = None,
     ):
         self.process: mp.Process = None
         self.start_time: Optional[float] = None
@@ -478,6 +480,11 @@ class PersistentAsyncCaller(AsyncCaller):
         self.cpu_priority = cpu_priority
         self.io_priority = io_priority
         self.cpu_shm_mode = cpu_shm_mode
+        # Plain dict of otel config + resource attributes (see async_ckpt/telemetry.py).
+        # spawn gives this process nothing from the parent's memory, so telemetry has
+        # to be bootstrapped fresh there from data passed explicitly through the
+        # ctx.Process args, the same channel as cpu_priority/io_priority above.
+        self.otel_bootstrap = otel_bootstrap
 
     def _start_worker(self, rank: int) -> None:
         """Start the background worker process.
@@ -506,6 +513,7 @@ class PersistentAsyncCaller(AsyncCaller):
                 self.cpu_priority,
                 self.io_priority,
                 self.cpu_shm_mode,
+                self.otel_bootstrap,
             ),
             daemon=self.background_worker_is_daemon,
         )
@@ -687,6 +695,7 @@ class PersistentAsyncCaller(AsyncCaller):
         cpu_priority: int = 10,
         io_priority: Optional[int] = None,
         cpu_shm_mode: bool = False,
+        otel_bootstrap: Optional[dict] = None,
     ):
         """Main function for the persistent checkpoint worker
 
@@ -714,6 +723,11 @@ class PersistentAsyncCaller(AsyncCaller):
                                         I/O priority unchanged. Use 3 (idle) to deprioritize
                                         checkpoint I/O. NOTE: class 1 = realtime (highest
                                         priority — NOT recommended for checkpoint workers).
+            otel_bootstrap (dict, Optional): otel config + resource attributes built by the
+                                             caller and threaded through via ctx.Process args,
+                                             since this process is spawned fresh and inherits
+                                             nothing from the parent. None/empty means telemetry
+                                             stays disabled.
 
         """
         # Align library loggers in this process without mutating the root logger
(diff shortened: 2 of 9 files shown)


hexinw-nvidia:
## What this PR is

Effectively Russell's #401, taken over while he is out, squashed into one commit with him as co-author. The telemetry itself — what is instrumented, the span names, the attributes, the restart-cycle model — is his work and is unchanged. What is added here is a restructure of *where* that telemetry lives and how it attaches to the code it instruments.

## What the restructure changes

`shared_utils` owns only the nemo-lens seam. Everything that decides *what* to instrument lives with the subsystem doing it:

```
shared_utils/telemetry.py                34 code   the seam (HAS_NEMO_LENS + helpers)
fault_tolerance/telemetry.py            455 code   span tree + recorder + agent facade
checkpointing/async_ckpt/telemetry.py    92 code   checkpoint worker facade
shared_utils/profiling.py               +57        event record, observer slot, phase CM
```

The recorder *observes* the fault-tolerance profiler rather than being driven by the launcher, so `profiling.py` knows nothing about OTel and adding an event means adding a row to one table.

The point of this is the call-site surface, which is what keeps future telemetry iterations out of the instrumented modules:

```
launcher.py   6 semantic calls   annotate_cycle · stage_outcome ×2 · finish_cycle · worker_env · shutdown
core.py       4 semantic calls   request_span · write_span · shutdown
```

Telemetry work after this should land in `fault_tolerance/telemetry.py` or `checkpointing/async_ckpt/telemetry.py`, not in `launcher.py` or `core.py`. We expect a couple more iterations on the telemetry itself, and this is the shape that makes them cheap.

Also from review: no no-op stubs for a missing nemo-lens (a stub has to fake a tracer, which needs opentelemetry, which may also be absent — so it could raise from the path meant to be the safe one; callers check `HAS_NEMO_LENS` once instead).

## Behaviour vs #401

Verified rather than asserted: both implementations driven through the same 14-event sequence — full cycle, failure, teardown, standby wait, promotion, eviction, with cold-start anchors and a staged outcome — and every emitted span diffed.

```
#401 emitted 15 spans, #402 emitted 15
IDENTICAL — same names, timestamps and attributes, in the same order
```

Two deliberate deviations, both outside that: a checkpoint-worker telemetry setup failure no longer propagates (it would previously take down the worker), and the agent clears its profiling observer on shutdown.

Tests: 37 covering the span-tree state machine — full cycle, phase sweep, evicted node, standby/await, staged outcomes, cold start, flush points, profiler wiring, and the failed-setup path. They run without opentelemetry or nemo-lens installed, since the recorder only needs an object with `start_span()`.

## On the Greptile finding

Greptile is right that a queue adopting the pre-warmed checkpoint worker inherits that worker's telemetry configuration, because `otel_bootstrap` is fixed at spawn.

The mechanism causing it predates this PR: `_get_async_caller()` adopts the pre-warmed caller wholesale, discarding the consuming queue's values for **all** of `is_daemon`, `cpu_priority`, `io_priority`, `sigterm_timeout`, `cpu_shm_mode` and `otel_bootstrap`. Telemetry is simply the newest parameter to inherit an existing hazard, not a new one.

Worth noting on severity: nothing in NVRx calls `warmup_persistent_caller()` — it is a public API for Megatron/NeMo only — so this is not reachable from any code path in this repo. It requires a consumer to pass `otel_bootstrap` to `AsyncCallsQueue(...)` but not to `warmup_persistent_caller(...)`, i.e. a partial integration.

Handled here to the extent this PR owns it:
- when the pre-warmed process is dead, a fresh worker is spawned anyway, so the queue's `otel_bootstrap` is adopted — correctness at no cost;
- when it is alive and the configurations differ, a warning naming the PID and pointing at `warmup_persistent_caller()`, so a mis-integration is visible rather than silently emitting no spans.

Deliberately **not** done here: discarding a live pre-warmed worker to respawn it with the right configuration. That fixes telemetry by throwing away the warm-up latency benefit, which is a checkpointing trade-off rather than a telemetry one.

The general fix — `_get_async_caller()` validating the pre-warmed caller's full configuration against the consuming queue's, and either respawning or refusing — belongs in a separate PR against the checkpointing owners. `cpu_shm_mode` is the one I would prioritise there: it controls whether the worker initialises a CUDA device, so a mismatch either leaves the worker holding a CUDA context it was meant to avoid, or risks the `pidfd_getfd` handle-type mismatch that the dummy-tensor allocation exists to prevent. Unlike the telemetry case, that one needs no consumer error to bite.

## #401

Superseded by this PR. Happy to keep it open until Russell is back if he would rather land his own.

PR #404, opened 2026-08-11
Move optional attribution deps behind attribution extra

namitdhameja:
## Summary

Attribution currently pulls in a large dependency stack by default, including LogSage, LangChain, MCP, and Slack-related packages. That makes the default NVRx install heavier than necessary for users who only need the regular FT / Restart Agent / attrsvc path.

This PR moves the optional attribution dependency stack behind the `nvidia-resiliency-ext[attribution]` extra. The functionality remains available for users who need those attribution workflows, while the default install stays smaller and easier for regular NVRx users.

## What changed

- Keep `nvrx-attrsvc` and the direct Restart Agent backend in the default install.
- Move optional attribution dependencies to `[attribution]`:
  - `logsage`
  - `langchain-core`
  - `langchain-openai`
  - `mcp`
  - `setproctitle`
  - Slack SDK packages
- Remove MCP as a launcher-managed / attrsvc-selectable backend.
- Keep lower-level MCP code available for explicit use after installing `[attribution]`.
- Update docs, service scripts, and tests for the new install contract.


hexinw-nvidia (approved):
Cool!

PR #406, opened 2026-08-16
NemoLens integration

shurkat-nvidia:
Integrates in NemoLens through lightweight additions

 - Adds a telemetry module in NVRX
 - Adds minimal with blocks and decorators for OTel spans
 - Allows spans to be opened and closed across function or module boundaries
 - Adds design documentation for the integration


src/nvidia_resiliency_ext/fault_tolerance/launcher.py
```diff
@@ -654,6 +659,13 @@ def _invoke_run_with_any_failed_policy(self, role: str = DEFAULT_ROLE) -> RunRes
 
         logger.info("[%s] starting workers for entrypoint: %s", role, spec.get_entrypoint_name())
 
+        # Every agent exports, so dl.rank / service.instance.id have to differ per node.
+        # Rendezvous has not assigned elastic ranks yet, hence the infrastructure rank.
+        self._tel_handle = telemetry.setup_telemetry(
+            max(get_infrastructure_rank(), 0), int(os.environ.get("SLURM_NNODES", "1"))
+        )
+        atexit.register(self._tel_handle.shutdown)
```
hexinw-nvidia: atexit runs at Python interpreter finalization, after sys.exit unwinds. It can stall ft_launcher exit at the worst case. I think the previous implementation of leaving the telemetry shutdown in the finally statement of run() is better.

hexinw-nvidia (commented):
Nice work. This is a much cleaner design than PR#402. I only have one comment on the atexit change. Thanks.

pyproject.toml
```diff
@@ -51,6 +51,7 @@ httpx = ">=0.24.0"
 langchain-core = { version = ">=0.3.51,<0.4.0", optional = true }
 langchain-openai = { version = ">=0.3.0,<1.0.0", optional = true }
 mcp = { version = ">=1.15.0,<2.0.0", optional = true }
+nemo-lens = { version = ">=0.2.0", extras = ["sdk"], optional = true }
```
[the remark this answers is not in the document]
shurkat-nvidia: This is fine. nemo-lens 0.2.0 will be released when the next nvrx release is made. nemo-lens 0.1.0 does not have the APIs necessary for this change.

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -2700,20 +2703,25 @@ def _perform_rendezvous(self) -> None:
 
         def pre_join_hook() -> None:
             health_check_start = time.monotonic()
+            record_event("health_check.started")
```
rhewett-nv: Can we make this "end the previous span" and then "start this span"?

src/nvidia_resiliency_ext/fault_tolerance/launcher.py
shurkat-nvidia: Addressed in dfeb15e.

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
shurkat-nvidia: Addressed in dfeb15e. Dropped events entirely.

hexinw-nvidia (commented):
Looks good to me. Thanks for addressing the feedback.

docs/design/telemetry/NEMO_LENS.md
```diff
@@ -0,0 +1,207 @@
...
+
+Every launcher agent exports. Each node runs its own collector and we want per-node visibility, so the shim defaults `export_strategy` to `all_ranks` rather than relying on nemo-lens's `single_rank` default. An explicit `NEMO_LENS_EXPORT_STRATEGY` still wins, so volume stays tunable.
+
+The launcher's identity is resolved before rendezvous, where no elastic rank exists yet: `get_infrastructure_rank()` (NVRx's stable node identity) and `SLURM_NNODES`, each falling back to `0` / `1`. The elastic `group_rank` is set on the cycle span once rendezvous assigns it.
+
+## Fault Tolerance Span Lifecycle
+
+### Everything is a span
+
+There are no OTel events. Every phase that was once a start/end event pair is a span, so its duration is a first-class queryable value rather than something a backend has to reconstruct by subtracting two timestamps.
```
rhewett-nv: Claude, you didn't need to write this...

docs/design/telemetry/NEMO_LENS.md
```diff
@@ -0,0 +1,207 @@
...
+
+| Shape | Mechanism | Spans |
+| ----- | --------- | ----- |
+| The span *is* a method | `@traced` decorator | `worker_start`, `teardown` |
+| The span is a block | `with managed_span(...)` | `round_wait`, `health_check`, both `ckpt` spans |
+| The span outlives its block | `ManualSpan` | `cycle`, `rendezvous` |
+
+Using `@traced` for the first group means the instrumented method bodies are untouched.
+
+`ManualSpan` is needed for exactly two spans. A `cycle` opens at rendezvous and ends only after teardown. A `rendezvous` round opens after the round has opened — deliberately excluding the hot-spare wait, which would otherwise dominate the measurement — and ends either when the node is assigned an active rank or when the next round begins.
```
rhewett-nv: ManualSpan covers the case where the span open/close crosses code block boundaries

docs/design/telemetry/NEMO_LENS.md
```diff
@@ -0,0 +1,207 @@
...
+
+An exclusion needs no separate signal: the `UnhealthyNodeException` propagating out of the `health_check` span is recorded by nemo-lens as `StatusCode.ERROR` with the reason as the description, plus an `exception` event carrying the type and message.
+
+### Initialization
+
+`setup_telemetry` is called once per launcher-agent process at the top of `run()`, before the first rendezvous, and `shutdown()` is called from that method's existing `finally`. Deliberately not `atexit`: that runs at interpreter finalization, after `sys.exit` has unwound, where a flush to an unreachable collector would stall `ft_launcher` exit.
```
rhewett-nv: This is true for ft_launcher.  what about when used for async ckpt?

docs/design/telemetry/NEMO_LENS.md
```diff
@@ -0,0 +1,207 @@
...
+
+nemo-lens's `_OpenSpanCloser` closes any span still open at that point, marking it `nemo.span.truncated`. NVRx relies on it for exactly one case — a cycle interrupted by a signal — rather than wrapping the monitor loop in a `try/finally` purely to relabel that span. A truncated cycle with no `nvrx.cycle_outcome` is a faithful description of what happened.
+
+There is no per-cycle `force_flush`. Spans are exported by the batch processor on its own schedule while the agent keeps running, and flushed by `shutdown()` on exit; flushing synchronously on the restart path would only add latency to recovery.
```
rhewett-nv: OK?

src/nvidia_resiliency_ext/fault_tolerance/launcher.py
```diff
@@ -450,6 +453,7 @@ def __init__(
         self._children_pgids: Set[int] = set()
         self._restart_policy = restart_policy
         self._node_id = self._get_fq_hostname()
+        self._cycle = telemetry.ManualSpan()
```
rhewett-nv: Maybe name something lime `_cycle_span` so it is clear that this is the span object being opened/closed, not the cycle itself

docs/design/telemetry/NEMO_LENS.md
shurkat-nvidia: I agree. Cleaned up.

docs/design/telemetry/NEMO_LENS.md
shurkat-nvidia: It's in `handle_sigterm`. Do you want this added to the doc?

docs/design/telemetry/NEMO_LENS.md
shurkat-nvidia: Cleaned up, I assume that was the intention?

src/nvidia_resiliency_ext/fault_tolerance/launcher.py
shurkat-nvidia: Done, also renamed to `run_span`

src/nvidia_resiliency_ext/fault_tolerance/launcher.py
```diff
@@ -1380,6 +1457,40 @@ def any_rank_failed(self) -> bool:
...
+            "nvrx.max_restarts": spec.max_restarts,
+            "nvrx.remaining_restarts": self._remaining_restarts,
+        }
+        try:
+            run_id = spec.rdzv_handler.get_run_id()
+        except Exception:
+            logger.debug("Rendezvous run id unavailable for telemetry", exc_info=True)
+        else:
+            if run_id is not None:
+                attrs["nvrx.rdzv_run_id"] = run_id
```
hexinw-nvidia: The try...except is not needed. spec.rdzv_handler.get_run_id() should always be valid.
Also, the "run_id" will not be None. Refer to the init check in the ft_rendezvous_barrier.py

2433         if not settings.run_id:
2434             raise ValueError("The run id must be a non-empty string.")

src/nvidia_resiliency_ext/fault_tolerance/launcher.py
```diff
@@ -450,6 +460,8 @@ def __init__(
         self._children_pgids: Set[int] = set()
         self._restart_policy = restart_policy
         self._node_id = self._get_fq_hostname()
+        self._cycle_phase = telemetry.Phase()
+        self._run_phase = telemetry.Phase()
```
hexinw-nvidia, line 465: What is the difference between "_cycle_phase" and "_run_phase"?

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -1903,7 +1906,15 @@ def perform_rendezvous(
             # Hot spares and late-arriving nodes wait here indefinitely until a failure
             # opens the next round. Also checks for permanent shutdown.
             # Note: _wait_for_rendezvous_open() raises RendezvousGracefulExitError on shutdown.
-            self._wait_for_rendezvous_open(node_desc)
+            # Closed before the wait, so a hot spare idling here does not sit inside the
+            # previous round's span.
+            self._rdzv_span.close()
```
hexinw-nvidia, line 1912: Is this still required? The try...finally block around the perform_rendezvous() already guarantees the span is closed.

src/nvidia_resiliency_ext/fault_tolerance/launcher.py
```diff
@@ -1076,6 +1134,24 @@ def _start_workers(self, worker_group: WorkerGroup) -> Dict[int, Any]:
...
+
+        # Resource attributes for the worker, span attributes for the agent: a worker
+        # process is new each cycle, the agent outlives them.
+        worker_resource_attrs = {"nvrx.cycle": restart_count, "nvrx.membership": "active"}
+        try:
+            worker_resource_attrs["nvrx.infra_rank"] = get_infrastructure_rank(
+                skip_nodename_logic=True
+            )
+        except Exception:
+            logger.debug("Infrastructure rank unavailable for worker env", exc_info=True)
```
hexinw-nvidia: What exception is it catching? Can we be explicit about the catch?

src/nvidia_resiliency_ext/fault_tolerance/launcher.py
shurkat-nvidia: `cycle` is the parent, `run` is the nested span that corresponds to running Megatron.

`cycle` will include health check, rendezvous, etc.

src/nvidia_resiliency_ext/fault_tolerance/launcher.py
shurkat-nvidia: Yeah, removed this.

src/nvidia_resiliency_ext/fault_tolerance/launcher.py
shurkat-nvidia: Changed to only catch `ValueError` or `RuntimeError`.

src/nvidia_resiliency_ext/fault_tolerance/launcher.py
```diff
@@ -450,6 +460,17 @@ def __init__(
...
         self._node_id = self._get_fq_hostname()
+        # Nested, not alternatives. The cycle opens at rendezvous and closes when the
+        # next one opens, so it covers rendezvous, the health check, worker launch,
+        # the run itself and teardown. The run opens only once _start_workers has
+        # returned and closes the moment the workers stop, so it covers just the
+        # workload executing.
+        #
+        # The difference is the point: cycle minus run is this cycle's resiliency
+        # overhead -- the time the job spent getting ready to train rather than
+        # training. Collapsing them into one span would make that unanswerable.
```
shurkat-nvidia: Ugh, need to remove these novellas

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
shurkat-nvidia: The idea here is that we close the span inside the rendezvous function within each loop iteration. If the function terminates, we close the last span in the `finally` block

hexinw-nvidia: Right. This is for the hot-spare and latecomer.

hexinw-nvidia:
Looks like the GitLab CI failed.

docs/design/telemetry/NEMO_LENS.md
shurkat-nvidia: Deleted manual span entirely.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -57,6 +57,7 @@
 except ImportError:
     HAVE_PSUTIL = False
 
+from ...shared_utils import telemetry
```
ankurv-nvidia: (NIT: Feel free to keep out of scope.  Mostly wondering what sort of linter rules do we enforce in the CI by default ?) 

lint : relative-imports (TID252)
https://docs.astral.sh/ruff/rules/relative-imports/

I dont have a strong opinion on from "."  or but 3 levels of relative parent directory imports sounds like something we should avoid

src/nvidia_resiliency_ext/shared_utils/telemetry.py
```diff
@@ -0,0 +1,402 @@
...
+import threading
+import time
+from collections.abc import Callable, Iterator, Mapping, MutableMapping
+from contextlib import contextmanager, nullcontext
+from typing import TYPE_CHECKING, Any, Optional, ParamSpec, TypeVar
+
+if TYPE_CHECKING:
+    from contextvars import Token
+
+    from nemo.lens import TelemetryHandle
```
ankurv-nvidia: Wont this throw if nemo.lens isnt installed 
Given that its under TYPE_CHECKING i am assuming this doesnt really affect actual runs but does this mean developer scenarios where nemo.lens isnt available would trip off linters or editor-intellisense if nemo.lens isnt installed ?

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
```diff
@@ -362,19 +363,21 @@ def separate_for_gpu_ipc(items, resolved_data, dequantized_flags):
                     # the worker for checkpoint N may still be reading these buffers
                     # while we are about to overwrite them with checkpoint N+1 values.
                     if FileSystemWriterAsync._shm_drain_callback is not None:
-                        FileSystemWriterAsync._shm_drain_callback()
+                        with telemetry.span("nvrx.ckpt.phases", "nv.nvrx.ckpt.save.shm_drain"):
```
ankurv-nvidia: Any reason to have this group-name vs span-name asymmetry  ?
missing nv prefix in the group and no "phases" in the name ?
Would be nice to keep them aligned so its easier to turn on / off groups based on span names in the telemetry

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -668,6 +669,11 @@ def __init__(
             # not just key existence, to decide whether rendezvous is shut down.
             self.store.set(self.shutdown_key, b"")
 
+    @property
+    def domain_id(self) -> Optional[str]:
+        """Return the cached domain ID, or None when unavailable."""
+        return None if self._cached_domain_id == "none" else self._cached_domain_id
```
ankurv-nvidia, line 676: Is this really supposed to contain the value "none" for none ?
given `self._cached_domain_id` is typed  `Optional[str]` should this be ?

```suggestion
        return None if self._cached_domain_id is None or self._cached_domain_id == "none" else self._cached_domain_id
```

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -1903,7 +1910,24 @@ def perform_rendezvous(
             # Hot spares and late-arriving nodes wait here indefinitely until a failure
             # opens the next round. Also checks for permanent shutdown.
             # Note: _wait_for_rendezvous_open() raises RendezvousGracefulExitError on shutdown.
-            self._wait_for_rendezvous_open(node_desc)
+            #
+            # Close the previous cycle after its rendezvous scope has exited.
+            # The round may change while the next wait is active.
+            if self._agent is not None:
```
ankurv-nvidia, line 1917: On the first entry in the while loop is there anything to close ?
Should this be 
```suggestion
            if self._agent is not None and cycle_attributes is not None:
```

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -1903,7 +1910,24 @@ def perform_rendezvous(
...
+                self._agent.close_telemetry_cycle(cycle_attributes)
+            cycle_attributes = None
```
ankurv-nvidia, line 1919: NIT: 
Not sure if the above "first-loop-extraneous-close" issue is a real issue and depending on how you want to handle it . It might be worth adding
```
            cycle_attributes = {
                    "nv.nvrx.cycle.outcome": "unknown",
                    "nv.nvrx.ftl.membership": "unknown",
                }
```
So we're always guaranteed a cycle_attribute once a telemetry cycle is opened and someone accidetally doesnt continue the loop without setting one. ..

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
```diff
@@ -1915,136 +1939,158 @@ def perform_rendezvous(
...
+                # The store host set round_done_key=1 only AFTER writing all rank keys, so
+                # any participant that sees round_done=1 can immediately read its rank.
+                rank, total_participants = self._wait_for_round_done(node_desc, rank_key)
 
-            if rank != GroupRankStatus.UNASSIGNED.value and rank < min_nodes:
-                # Active rank: return to launcher to start training workers.
-                return rank, total_participants
+                if rank != GroupRankStatus.UNASSIGNED.value and rank < min_nodes:
+                    # Active rank: return to launcher to start training workers.
+                    if rdzv_span is not None:
```
ankurv-nvidia: could this really be none given that we're in a span ?

src/nvidia_resiliency_ext/shared_utils/telemetry.py
```diff
@@ -0,0 +1,402 @@
...
+from __future__ import annotations
+
+import functools
+import logging
+import os
+import threading
+import time
+from collections.abc import Callable, Iterator, Mapping, MutableMapping
+from contextlib import contextmanager, nullcontext
+from typing import TYPE_CHECKING, Any, Optional, ParamSpec, TypeVar
```
ankurv-nvidia: NIT: 
Given that minimum python version we support is 3.10 and that this is a new file . It'd be nice to modernize the typing to be `str | None` instead of `Optional[str]` etc.

ankurv-nvidia: as a low hanging fix 
can you run `ruff check -fix` on this file 
Vscode shows lots ruff linter warnings, most of them around use `Optional` instead of ` | None`

src/nvidia_resiliency_ext/shared_utils/semconv.py
```diff
@@ -0,0 +1,22 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Telemetry names NVRx shares with the framework that drives it.
+"""
+
+#: Queue-assigned index correlating schedule, request and finalize spans.
+#: Returned by schedule_async_request for frameworks to record on their own spans.
+#: Indices are local to a queue; ambient context determines trace boundaries.
+CKPT_CALL_IDX = "nv.nvrx.ckpt.call_idx"
```
ankurv-nvidia, line 31: A wholly separate file for a single variable seems a little overkill. 
Any reason for it to be this way ?

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
shurkat-nvidia: If `self._cached_domain_id` is `None`, then the previous code returns `None` anyway, and avoids an `is None` check. The code only has to convert `"none"` to `None`.

src/nvidia_resiliency_ext/shared_utils/semconv.py
shurkat-nvidia: Yes, because this file is meant to be used by NVRX users. `telemetry.py` is meant to be internal to NVRX.

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
shurkat-nvidia: Close is idempotent, so leaving in the close unguarded is fine. If attributes is None, it doesn't update the attributes on the span (as in, there are no additional attributes to be added), whereas listing these as "unknown" will add the attributes, potentially causing noise.

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
shurkat-nvidia: No, the attributes None check is handled in telemetry.py.

src/nvidia_resiliency_ext/shared_utils/semconv.py
shurkat-nvidia: Added in span group names also here, so we now have more than one constant :D

src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py
shurkat-nvidia: In the version you reviewed, it could be, if nemo lens wasn't installed. This technically leaked a lens availability check into business code, so I refactored it to remove the not none check.

src/nvidia_resiliency_ext/shared_utils/telemetry.py
shurkat-nvidia: Done. Fixing everything is out of scope, but it should be better now.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
shurkat-nvidia: Changed it so that the group and span prefixes are consistent now. Also changed to using named constants for group names.

src/nvidia_resiliency_ext/shared_utils/telemetry.py
shurkat-nvidia: Yeah, was a bit ugly. Changed it so mypy now ignores modules it can't import.

src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
shurkat-nvidia: Changed to an absolute import.

Issue #405, opened 2026-08-14
performance improvements during cpu-shm snapshotting


ankurvdev:
in filesystem_async.py::prepare_write_data
```
for tensor in cacheable_data:
     contiguous_tensor = tensor.contiguous()
     shm_t = torch.empty(
        contiguous_tensor.shape, dtype=contiguous_tensor.dtype
     ).share_memory_()
```
The contiguous potentially causes double copy of the memory and can be potentially omitted for better performance.

In the same function
torch.cuda.synchronize can be replaced with a per-stream event based synchronize to avoid forcing all streams to be synchronize to avoid a synchronize perf penalty

0z5a:
I opened draft PR #407 for the two optimizations described here. It is stacked on #376 so that the diff only contains the direct-copy and current-stream event changes.

On one H200 NVL, the event wait returned in 2.564 ms versus 1124.072 ms for device-wide synchronization while an unrelated stream was active. Direct copy reduced CPU non-contiguous snapshot-copy latency by 52-58% from 64 MiB through 1 GiB; CUDA copy latency was neutral on PyTorch 2.11 because `copy_` currently materializes the contiguous CUDA temporary internally.

The PR is marked WIP pending confirmation that this combined scope is appropriate.

PR #407, opened 2026-08-17
[WIP] #405 - Optimize CPU-SHM snapshot copies

0z5a:
## Summary

- copy CPU and CUDA sources directly into contiguous shared-memory snapshots instead of explicitly materializing `source.contiguous()` first;
- replace device-wide synchronization in both the initial-allocation and cache-reuse paths with an event recorded on the current CUDA stream;
- preserve the existing wait-after-eight-copies staging-page backpressure and the SHM cache drain/ownership protocol;
- cover CPU-only behavior, event scoping, and contiguous/transpose/strided-offset inputs on CPU and CUDA.

Refs #405.

## Dependency

This is a stacked WIP PR on top of #376 and therefore targets its upstream branch, `nvbug6439229`. Once #376 lands, this branch can be rebased and the PR base changed to `main`.

#376 is required because it routes CPU-resident tensors through the same owned CPU-SHM snapshot path. This PR does not change its snapshot isolation or reuse/drain semantics.

## Synchronization semantics

The event remains a host-side completion barrier for the copies that produce the snapshot, but it no longer waits for unrelated work on other CUDA streams. The periodic wait after eight CUDA sources is retained as backpressure for pageable/shared-memory staging pages.

## H200 benchmark

Environment: one NVIDIA H200 NVL, PyTorch `2.11.0+cu130`, 141 GiB `/dev/shm`; square FP32 transpose source; contiguous SHM destination. Copy-only numbers are medians of nine iterations after warmup.

### Reused SHM destination

| Source | Size | Current | This PR | Change |
|---|---:|---:|---:|---:|
| CPU | 64 MiB | 28.301 ms | 11.883 ms | -58.01% |
| CPU | 256 MiB | 99.407 ms | 47.185 ms | -52.53% |
| CPU | 1 GiB | 394.121 ms | 189.257 ms | -51.98% |
| CUDA | 64 MiB | 2.436 ms | 2.441 ms | +0.21% |
| CUDA | 256 MiB | 9.589 ms | 9.591 ms | +0.02% |
| CUDA | 1 GiB | 38.173 ms | 38.186 ms | +0.04% |

For fresh SHM allocation plus copy, the CPU-source path improved by 28-38% from 64 MiB through 1 GiB. The CUDA-source path remained within 0-3% in that three-iteration allocation-heavy microbenchmark.

PyTorch profiler shows that PyTorch 2.11 currently materializes a full-size contiguous CUDA temporary internally for a strided CUDA-to-CPU `copy_`, so CUDA peak temporary allocation remains unchanged. This PR removes the forced materialization from NVRx without regressing CUDA copy latency; the clear direct-copy benefit on this stack is for CPU-resident non-contiguous sources.

### Unrelated-stream interference

With `torch.cuda._sleep(2_000_000_000)` queued on an unrelated stream while copying a 64 MiB snapshot:

| Barrier | Host wait |
|---|---:|
| `torch.cuda.synchronize()` | 1124.072 ms |
| current-stream event | 2.564 ms |

The event wait returned while the unrelated stream was still running, and the SHM snapshot contents were correct.

## Validation

- `black --check .` (380 files unchanged)
- `isort --check-only .`
- `ruff check .`
- `tests/checkpointing/unit/test_preload_snapshot.py`: 10 passed
- `tests/checkpointing/unit/test_async_writer.py`: 19 passed on one H200/NCCL rank
- torn-checkpoint mode matrix `(cached structure on/off) × (CPU SHM on/off)`: 4 passed

### DIFF
--- a/src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
+++ b/src/nvidia_resiliency_ext/checkpointing/async_ckpt/filesystem_async.py
@@ -154,6 +154,13 @@ def _compute_tensor_data_ptrs(items: List[WriteItem], tensors: List[Any]) -> Tup
     return tuple(ptrs)
 
 
+def _synchronize_current_cuda_stream() -> None:
+    """Wait for work queued on the current CUDA stream without synchronizing the device."""
+    event = torch.cuda.Event()
+    event.record(torch.cuda.current_stream())
+    event.synchronize()
+
+
 @_disable_gc()
 def get_write_results_queue(mp_mode: str = 'spawn') -> mp.Queue:
     """Get or create a multiprocessing queue for write results.
@@ -371,10 +378,10 @@ class FileSystemWriterAsync(FileSystemWriter):
                         # pages (shm tensors are not pinned so copy_ is not a pinned DMA).
                         pending_cuda_copies += int(source_t.device.type == "cuda")
                         if pending_cuda_copies == 8:
-                            torch.cuda.synchronize()
+                            _synchronize_current_cuda_stream()
                             pending_cuda_copies = 0
                     if pending_cuda_copies:
-                        torch.cuda.synchronize()
+                        _synchronize_current_cuda_stream()
                     logger.debug(
                         f"Copied {len(shm_tensors)} tensors into reused shm allocations "
                         f"(key={key})"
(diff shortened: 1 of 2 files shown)


0z5a:
Hi @hexinw-nvidia Would you mind having a review?

PR #411, opened 2026-08-28
Category-driven STOP context in restart_agent

continue-revolution:
## Summary

Layers the 38-category L1 taxonomy and the `l1_category_confirmed_stop` policy context from PR #390 on top of PR #400's typed-claim architecture. The category signal runs last in the policy cascade, subordinate to PR #400's four deterministic classifier contexts (`workload_unrecoverable`, `concrete_confirmation_retry`, `workload_confirmation_retry`, `general_retry`) — it only fires when the primary is grounded and the picked category's `decision` is `STOP`.

## Why

Empirically, LLMs are ~30pp better at concrete classification (picking one of 38 named categories) than at abstract downstream reasoning (naming failure_domain and retry_outlook via prose). PR #400's classifier cascade relies on that abstract reasoning to hit `workload_unrecoverable`, which leaves a cluster of deterministic-workload STOP cases uncovered. This PR routes the concrete signal directly into L4 as an additional PolicyContext without touching PR #400's cascade order.

## What's in this PR

- **`l1/categories.py`** — 38-entry taxonomy (id, name, description, `decision` ∈ {STOP, RESTART}, `failure_domain`, `retry_outlook`)
- **`l1/response_contract.py`** — declares `category_selection` as a required top-level L1 field (OpenAI structured-output would otherwise silently skip an optional field on ~40% of calls)
- **`l1/prompts.py`** — renders the taxonomy into the L1 prompt and asks the model to pick one
- **`l1/validation.py`** — validates category_selection; five in-place normalizers ported from PR #390 for contract-repair fallbacks
- **`l1/openai_compatible.py`** — wires `repair_model_evidence` before schema validation
- **`l1/contracts.py`** — adds `category_selection()` accessor
- **`l4/policy.py`** — adds `l1_category_confirmed_stop` PolicyContext. Fires with `allowed_retries=0` when the primary failure is grounded and the picked category has `decision=\"STOP\"`. Runs after every deterministic classifier context, so it never overrides them.
- **`decision_pipeline.py`** — passes `l1_category_selection` into `L4CyclePolicyInput`
- **`docs/design/attribution/restart_agent/spike_l1_category_integration.md`** — design note explaining rationale and DECISIONS.md alignment
- Also flips categories 25 (\"CUDA OOM\") and 27 (\"CUDA OOM during checkpoint async-save\") from RESTART to STOP after product-policy discussion — a CUDA OOM under an unchanged workload is highly likely to reproduce on retry, and the retry cost is high

## Numbers

All numbers on 79-case subset_batch1 with `gold_v3` (CUDA OOM = STOP).

### 1. Final decision accuracy

| config | gpt | qwen397b | gemini | nemotron |
|---|---|---|---|---|
| PR #400 baseline | 69.6% | 75.9% | 93.7% | 82.3% |
| PR #390 (previous approach) | 92.4% | 89.9% | 92.4% | 87.3% |
| **This PR (spike v5)** | **100.0%** | **98.7%** | **96.2%** | **89.9%** |

Δ vs PR #400 baseline: **+7.6 to +30.4 pp**. Δ vs PR #390: **+2.6 to +7.6 pp**.

### 2. L1 category selection accuracy

| | gpt | qwen397b | gemini | nemotron |
|---|---|---|---|---|
| picked cat == gold cat | 94.8% | 88.6% | 92.3% | 78.2% |

### 3. L4 policy-question accuracy (failure_domain + retry_outlook)

Scored against the `recovery_assessment_expectation` in gold_v3.

| field | gpt | qwen397b | gemini | nemotron |
|---|---|---|---|---|
| failure_domain (workload/infrastructure) | 67.1% | 69.6% | 82.1% | 62.0% |
| retry_outlook (may_recover / cannot_recover) | 83.5% | 91.1% | 79.5% | 70.9% |

### 4. Natural-language root cause accuracy (LLM-judged)

Judge is GPT-5.5 comparing each model's `l1_assessment.root_cause_assessment.summary` + `primary_failure.failure_identity` against the human-authored `root_cause` sentence from `nemotron6_author_seed_labels.csv`.

**Lenient rubric** — mechanism-family match, minor detail slips acceptable:

| config | gpt | qwen397b | gemini | nemotron |
|---|---|---|---|---|
| This PR (spike v5) | 93.0% | 91.8% | 90.3% | 88.9% |
| PR #390 (previous approach) | 100.0% | 97.4% | 90.7% | 92.2% |

**Gaokao rubric** — same-meaning match (Chinese Gaokao reading-comprehension style: same event/subsystem/phase; paraphrase OK; wrong-mechanism WRONG even if family matches):

| config | gpt | qwen397b | gemini | nemotron |
|---|---|---|---|---|
| This PR (spike v5) | 65.8% | 68.4% | 67.1% | 67.1% |
| PR #390 (previous approach) | 71.6% | 69.9% | 63.5% | 64.1% |

## Interpretation

- The category signal is the load-bearing element. Where the picker is right, the decision is right. Where it's wrong, the decision is wrong ~95% of the time.
- The `l1_category_confirmed_stop` context fires on 24–28 cases per model with 82–100% precision (nemotron 82%, other three ≥ 96%).
- Reasoning-block accuracy (failure_domain, retry_outlook, NL root cause) is materially lower than category accuracy across every model. The pipeline routes around this gap by making the concrete-classification signal first-class policy input.
- Under the Gaokao rubric, PR #390 and spike v5 are within ±4pp of each other on every model. NL RC quality tracks the underlying LLM's ceiling, not the pipeline architecture.

## DECISIONS.md alignment

- **D2** (L2 non-overriding): the category context runs *after* every classifier context. Cannot flip a STOP from `workload_unrecoverable` to RESTART; cannot flip a RESTART from `port_bind_confirmation_retry` to STOP.
- **D8** (score-free deterministic): no scoring or confidence gate. The context fires purely on a category's typed `decision` field.
- **D9** (config-owned policy): the category → decision mapping lives in `l1/categories.py`, editable without code changes to L4.
- **D14** (separate surface from root): the category is a compact summary of the failure family, not a root-cause claim. Root cause continues to come from `l1_assessment.root_cause_assessment`.

## Follow-ups

- **NL root cause improvement** is not on the shortlist. Under Gaokao grading, both PR #390 and this PR land at 65–72% across models, indicating this is closer to the underlying LLM ceiling than a fixable pipeline gap. ~15% of the gap comes from gold labels that reference off-log context the analyzer cannot access (documented in `v4-key-lessons` design note). Higher-ROI targets: better nemotron category calibration; ports of PR #390's small deterministic gates that don't rely on category classification.
- The eval scorer at `tools/restart_agent_eval/src/restart_agent_eval/evaluate.py` reads `analyzer_trace.l1.parsed_evidence.model_recovery_assessment` — PR #400 renamed that path to `.semantic_payload.model_recovery_assessment`. Every recovery-block metric in results.jsonl reports 0% as a result. Fixing the scorer path is a small, high-value follow-up unrelated to this PR.

## Test plan
- [ ] `pytest tests/attribution/unit/` — all green
- [ ] Full spike v5 eval reproduces numbers reported above
- [ ] Manual review of one confirmed_stop fire and one deterministic-context fire per model to verify precedence

## Follow-up landed: opt-in two-way flip (`l1_category_confirmed_restart`)

Commit `a942ad4` adds a companion RESTART context — the mirror image of the STOP context. Disabled by default. When enabled via `policy_contexts.l1_category_confirmed_restart.enabled = True`, it overrides a `workload_unrecoverable` base_rule when:

1. The L1 category picker chose a RESTART-labeled taxonomy entry.
2. History shows zero prior attempts with the same `root_fingerprint` (first-occurrence guard).

Emits `workload_confirmation_retry` with `history_match_scope=ROOT_ONLY` and `allowed_retries=1`. On any recurrence, the guard suppresses the override; base_rule + exhaustion produce STOP. Runs after every deterministic classifier context and after the STOP category context.

**Why this exists.** On the 79-case eval, gemini and nemotron's abstract L1 signals (`retry_outlook=cannot_recover`, `established_by_current_log`) are over-confident on cases where the concrete category picker correctly identifies a RESTART family (grad-inf, progress-log assertion, CPU OOM). PR #400's base_rule cascade believes the abstract signals and lands STOP. This context restores the RESTART on first occurrence.

**Two-way eval on gold_v3, 79 cases × 4 models:**

| model | one-way | two-way | Δ |
|---|---|---|---|
| gpt | 100.0% | 98.7% | −1.3pp (LLM stochasticity on b1-048; new context inactive here) |
| qwen397b | 94.9% | 96.2% | +1.3pp |
| gemini | 97.5% | **100.0%** | **+2.5pp** |
| nemotron | 91.1% | 92.4% | +1.3pp |

**New context precision:** 5 fires, 5 correct, **100%**.

**Design guarantees:**

- **Only overrides `workload_unrecoverable`**, not other STOP-yielding rules. Minimal override surface.
- **First-occurrence guard.** On recurrence of the same `root_fingerprint`, context returns None and base_rule STOPs.
- **Secondary safety: `history_match_scope=ROOT_ONLY` + `allowed_retries=1`**. Even if the guard misses on a corner case, the `selected_policy_ledger` counts prior same-root attempts against the budget of 1 and produces `exhausted_by → STOP` on the second occurrence.
- **Config-gated.** `enabled=False` by default. Shipping code with no behavior change; operators opt in.

**Tests:** 7 new focused tests for the RESTART branch (first-occurrence positive, disabled-noop, recurrence-noop, STOP-cat-noop, RESTART-base-rule-noop, missing-cat-id-noop, cuda_oom-precedence). Full attribution/unit suite: **659 passed**.

Recommended review path: keep `enabled=False` in the default shipped config for now; flip on in a per-deployment override after a few operator runs.

### DIFF
--- a/src/nvidia_resiliency_ext/attribution/restart_agent/l1/categories.json
+++ b/src/nvidia_resiliency_ext/attribution/restart_agent/l1/categories.json
@@ -1,0 +1,233 @@
+{
+  "schema_version": "restart_agent_categories.v1",
+  "categories": [
+    {
+      "id": 1,
+      "name": "SLURM step cancelled / drain / preemption",
+      "description": "Clean external cancellation by the scheduler (scancel, drain, preemption); no application fault.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 2,
+      "name": "Node failure / SLURM externally terminated",
+      "description": "SLURM 'CANCELLED DUE TO NODE FAILURE' or other external node termination.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 3,
+      "name": "Off-file termination / external teardown (cause not in log)",
+      "description": "Healthy training then an external-signal all-thread faulthandler dump (no 'Fatal Python error' / no 'Current thread 0x'), log truncated mid-dump; no in-file CANCELLED / watchdog / IB / Xid / NVRx marker. A real termination occurred but its cause is off-file (most likely a SLURM cancel/preempt/timeout).",
+      "decision": "RESTART"
+    },
+    {
+      "id": 4,
+      "name": "No failure observed / clean planned exit",
+      "description": "Whole-file scan finds no failure: the run reached its exit_interval / final iteration, saved the final checkpoint, logged training energy, and only benign teardown (destroy_process_group / CudaIPC cleanup) follows. Includes the reset_opt_norm 'Unexpected keys' benign-WARNING trap and recovered-then-clean runs.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 5,
+      "name": "Incomplete / truncated log",
+      "description": "Whole-file scan finds no failure signal and the log has no clean ending; likely a log-collection failure.",
+      "decision": "EXCLUDED"
+    },
+    {
+      "id": 6,
+      "name": "NVRx used (multi-cycle, deferred)",
+      "description": "ft_launcher restarted the worker group in-job; one file concatenates multiple training cycles with separate root causes.",
+      "decision": "EXCLUDED"
+    },
+    {
+      "id": 7,
+      "name": "NVLink / GPU error",
+      "description": "Uncorrectable NVLink/GPU hardware error (Xid, ECC DBE, peer-access failure); the device fault may be transient and clear on an in-place restart.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 8,
+      "name": "IB port flap / NCCL watchdog timeout",
+      "description": "IB port error / RDMA retry exhaustion killing collectives; the port flap is typically transient and clears on an in-place restart.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 9,
+      "name": "Container / Pyxis setup failure",
+      "description": "Container never started (pyxis couldn't start container / enroot failure); transient node-local runtime vs bad image undecidable from one log.",
+      "decision": "EXCLUDED"
+    },
+    {
+      "id": 10,
+      "name": "Early/startup NCCL init or checkpoint-load timeout",
+      "description": "NCCL init / checkpoint-load collective timeout before the first completed training iteration; true cause usually not in this log.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 11,
+      "name": "NCCL timeout mid-training",
+      "description": "Watchdog collective timeout after at least one completed iteration; consequence-class symptom, upstream cause often absent from the log.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 12,
+      "name": "Distributed rendezvous / TCPStore socket timeout",
+      "description": "c10d/TCPStore rendezvous socket timeout as the INITIATING failure only. TCPStore socket errors appear as cascade signatures across many other failure modes: when any rank dies for any reason (CUDA OOM, SIGKILL, node failure, mid-training watchdog abort, checkpoint code error, etc.), the surviving ranks eventually time out waiting for the dead peer on the TCPStore, so many logs contain TCPStore timeout lines that are secondary. Pick this category ONLY when TCPStore rendezvous timeout is the primary observable event and there is no earlier crash, OOM, watchdog trigger, or other upstream fault in the log that better explains the failure. Otherwise pick the upstream category.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 13,
+      "name": "NCCL remote process exited / network error / bootstrap Message-truncated",
+      "description": "NCCL Error 6 (remote peer exited); infrastructure/peer failure. Also covers NCCL bootstrap/transport corruption ('Message truncated : received N bytes instead of M') during any distributed-checkpoint gather_object communicator setup - both LOAD-side startup and async-finalize save-side - where a downstream rank0 Segmentation fault in the same stack is NOT the root.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 14,
+      "name": "OSError Errno 98 / port bind race",
+      "description": "Fatal port-bind failure (EADDRINUSE) at the rendezvous master; startup race, job died at bind with no progress.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 15,
+      "name": "DataLoader worker I/O failure (Bus error, BrokenPipe, Lustre or NIC)",
+      "description": "Transient data-path I/O failure inside a DataLoader worker: Bus error, BrokenPipeError (Errno 108 transport endpoint shutdown) on Lustre-backed .bin readinto, /dev/shm exhaustion, or fatal Lustre read. Distinguishes from cat 16 by occurring in the DataLoader worker path.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 16,
+      "name": "Filesystem I/O error / Lustre or node-local FS / import visibility",
+      "description": "Transient filesystem error on a NON-checkpoint path: Lustre I/O stall (OSError Errno 5), missing dataset cache from filesystem-visibility lag, or ImportError/ModuleNotFoundError where the failing import path lives on a network-mounted code tree. Missing CHECKPOINT paths are cat 22 (STOP).",
+      "decision": "RESTART"
+    },
+    {
+      "id": 17,
+      "name": "Dataset missing indexed data files",
+      "description": "Dataset .idx/.bin files missing at dataset init; deterministic against the same config until dataset path/files are fixed.",
+      "decision": "STOP"
+    },
+    {
+      "id": 18,
+      "name": "PermissionError / dataset cache or filesystem permission",
+      "description": "Deterministic permission-denied (EACCES/Errno 13) on dataset cache or filesystem path.",
+      "decision": "STOP"
+    },
+    {
+      "id": 19,
+      "name": "Checkpoint failure / transient async-write I/O",
+      "description": "Transient I/O failure during the checkpoint async-save path on a checkpoint directory: filesystem_async 'Local process 0 encountered an error', PyTorch inline_container position mismatch, BrokenPipeError Errno 108 from a writer, OSError Errno 5 (Input/output error). Prefer this over cat 16 whenever the failing path is a checkpoint directory and the mechanism is on the write side, even if the surface is generic filesystem I/O. Load-side checkpoint I/O errors that do not fit cat 10 (early/startup timeout) route to cat 16 (Filesystem I/O) instead; they have the same decision (RESTART) but a different labeling story.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 20,
+      "name": "Checkpoint failure / corrupt checkpoint metadata on load",
+      "description": "Garbled checkpoint metadata buffer at load (UnpicklingError/UnicodeDecodeError); transient corruption, checkpoint likely reusable on healthy nodes.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 21,
+      "name": "Post-checkpoint progress-log assertion",
+      "description": "Assertion after checkpoint resume (progress-log bookkeeping); restart from last committed checkpoint is appropriate.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 22,
+      "name": "Checkpoint failure / missing checkpoint path",
+      "description": "Missing checkpoint path; cannot self-heal by restarting.",
+      "decision": "STOP"
+    },
+    {
+      "id": 23,
+      "name": "Checkpoint failure / deterministic checkpoint format or shape mismatch",
+      "description": "Checkpoint-load shape/key/TP-PP mismatch with zero completed iterations; deterministic against same checkpoint + code/config.",
+      "decision": "STOP"
+    },
+    {
+      "id": 24,
+      "name": "Checkpoint failure / async-save finalize code error",
+      "description": "Deterministic code error (e.g. NoneType AttributeError) in the post-training async-save FINALIZE path.",
+      "decision": "STOP"
+    },
+    {
+      "id": 25,
+      "name": "Checkpoint failure / save-path native crash (PyTorchStreamWriter)",
+      "description": "Native SIGSEGV in PyTorchStreamWriter::writeRecord during the first torch_dist checkpoint save; structural serializer buffer overrun deterministic against unchanged shape/config.",
+      "decision": "STOP"
+    },
+    {
+      "id": 26,
+      "name": "CUDA OOM during checkpoint async-save",
+      "description": "CUDA OOM inside the async checkpoint-save path on a near-full GPU. As with cat 32, an unchanged workload will keep hitting the same allocation ceiling on retry, so product policy calls this a STOP.",
+      "decision": "STOP"
+    },
+    {
+      "id": 27,
+      "name": "Bad token / gradient Inf or NaN",
+      "description": "Non-finite gradient/loss from a bad token or transient overflow; external guidance is restart.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 28,
+      "name": "Launch config or workload code deterministic error (TypeError/AttributeError)",
+      "description": "Deterministic runtime error inside already-loaded workload code that will recur on unchanged retry: TypeError/AttributeError on None value threaded through model code (e.g. float(None) in moe_layer postprocess), missing MASTER_ADDR, argparse error, or world-size mismatch. Excludes ModuleNotFoundError / ImportError where the failing import path resolves under Lustre or a mounted network filesystem - those are cat 16 (filesystem visibility), not a workload code fault, even if no adjacent FileNotFoundError is present on the same line.",
+      "decision": "STOP"
+    },
+    {
+      "id": 29,
+      "name": "Triton cubin / TorchInductor software toolchain error",
+      "description": "Triton autotuner KeyError 'Unknown key: cubin' on mamba_ssm kernels (idx 1311/1418/1420) + TorchInductor StaticCudaLauncher 'CUDA driver error: file not found' at JIT warmup (idx 1417); clean Python exceptions, multi-node, deterministic in-container.",
+      "decision": "STOP"
+    },
+    {
+      "id": 30,
+      "name": "CUDA illegal op in CUDA graph capture",
+      "description": "Illegal CUDA operation during graph capture; deterministic code bug on the same code path.",
+      "decision": "STOP"
+    },
+    {
+      "id": 31,
+      "name": "CPU OOM / Linux OOM killer",
+      "description": "Host memory exhausted (oom_kill). Old STOP rested on a same-job/config recurrence presumption, not on in-log evidence.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 32,
+      "name": "CUDA OOM",
+      "description": "Device OOM at fixed batch/config. A CUDA OOM under an unchanged workload is highly likely to reproduce on the next attempt; product policy calls this a STOP because the retry cost is high and the outcome is stable.",
+      "decision": "STOP"
+    },
+    {
+      "id": 33,
+      "name": "CUDA unspecified launch failure",
+      "description": "cudaErrorLaunchFailure; a single log cannot separate transient hardware/runtime state from a deterministic kernel/code bug.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 34,
+      "name": "Segmentation fault / native runtime crash",
+      "description": "Single-rank (rank 608) SIGSEGV in torch's pinned host allocator during GPTDataset index load, wild HIGH address (not 0x8), no self-reported corruption guard and no HW signal; a single log genuinely cannot separate a torch race from host-memory corruption.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 35,
+      "name": "Node-local native SIGBUS (Triton/TE native kernel)",
+      "description": "Native SIGBUS confined to one node/tray, two forms: (a) during Triton JIT .so dlopen/mmap at kernel load (idx 1006; driver.py compile_module_from_src -> create_module; zero iterations); (b) mid-training in the TE/Triton MoE permutation forward (idx 534; permutation.py; one node's 4 ranks after ckpt 27600 while thousands of peers keep running). Node-local hardware/memory fault.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 36,
+      "name": "Startup native SIGSEGV / torch pin_memory host-allocator",
+      "description": "Native SIGSEGV 'address not mapped to object at address 0x8' (null-ptr deref) with a byte-identical backtrace through at::native::_pin_memory into the libtorch_cuda CachingHostAllocator pinned-map operator[], fired at [before the start of training step] with zero completed iterations. pin_cpu_grads/pin_cpu_params=True forces the path in-log; crashing node sets differ across jobs (rules out node-local hardware); reproduces every rerun. 20 samples = 11 unique files (9 md5 mirror pairs).",
+      "decision": "STOP"
+    },
+    {
+      "id": 37,
+      "name": "Native heap corruption / glibc malloc abort",
+      "description": "Native glibc heap-integrity abort (\"malloc(): unaligned tcache chunk detected\") on a rank pre-first-iter; deterministic in-process heap corruption traceable to the initialization software stack, recurs on unchanged code/library/container image.",
+      "decision": "STOP"
+    },
+    {
+      "id": 38,
+      "name": "Suspected SDC / DP-replica weight-hash mismatch",
+      "description": "DP-replica weight-hash mismatch (suspected silent data corruption); restart from last good checkpoint is plausibly safe.",
+      "decision": "RESTART"
+    }
+  ]
+}
--- a/src/nvidia_resiliency_ext/attribution/restart_agent/l1/categories.py
+++ b/src/nvidia_resiliency_ext/attribution/restart_agent/l1/categories.py
@@ -1,0 +1,114 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Static failure-category catalog used by the L1 category-selection field.
+
+The 38 category entries live in ``categories.json`` next to this module.
+This module loads and validates the JSON at import time, exposing a frozen
+``CATEGORIES`` tuple and the ``category_by_id`` accessor. Keeping the data
+in JSON lets non-engineers extend the taxonomy without touching Python and
+makes the catalog easier to diff, review, and reuse from other tools.
+
+Loading is done via ``importlib.resources`` so the file is discoverable
+both from a source checkout and from an installed wheel.
+"""
+
+from __future__ import annotations
+
+import json
+from dataclasses import dataclass
+from importlib import resources
+from typing import Any
+
+CATEGORIES_SCHEMA_VERSION = "restart_agent_categories.v1"
+# Allowed values for CategoryDef.decision. "EXCLUDED" here means the category
+# is excluded from the eval-metric / policy-activation surface (e.g. no
+# failure observed, incomplete/truncated log, or a multi-cycle NVRx-managed
+# run). It is unrelated to SLURM/node exclusion.
+_ALLOWED_DECISIONS = frozenset({"STOP", "RESTART", "EXCLUDED"})
+_REQUIRED_FIELDS = ("id", "name", "description", "decision")
+
+
+@dataclass(frozen=True)
+class CategoryDef:
+    """One curated failure-category definition."""
+
+    id: int
+    name: str
+    description: str
+    decision: str  # STOP | RESTART | EXCLUDED
+
+
+def _load_categories() -> tuple[CategoryDef, ...]:
+    """Read and validate categories.json. Raises on any schema violation."""
+
+    with resources.files(__package__).joinpath("categories.json").open("r") as f:
+        payload: Any = json.load(f)
+
+    if not isinstance(payload, dict):
+        raise ValueError("categories.json must be a JSON object")
+    if payload.get("schema_version") != CATEGORIES_SCHEMA_VERSION:
+        raise ValueError(f"categories.json schema_version must be {CATEGORIES_SCHEMA_VERSION!r}")
+    raw = payload.get("categories")
+    if not isinstance(raw, list) or not raw:
+        raise ValueError("categories.json 'categories' must be a non-empty array")
+
+    entries: list[CategoryDef] = []
+    seen_ids: set[int] = set()
+    seen_names: set[str] = set()
+    for index, item in enumerate(raw):
+        if not isinstance(item, dict):
+            raise ValueError(f"categories[{index}] must be an object")
+        missing = [k for k in _REQUIRED_FIELDS if k not in item]
+        if missing:
+            raise ValueError(f"categories[{index}] missing required fields: {', '.join(missing)}")
+        cid = item["id"]
+        if not isinstance(cid, int) or isinstance(cid, bool) or cid < 1:
+            raise ValueError(f"categories[{index}].id must be a positive integer")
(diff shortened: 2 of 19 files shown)


continue-revolution:
Rebased onto latest \`agent/restart-agent-operational-logging\` HEAD (5097406) and marked ready. Rebase required two fixups (committed as \`64be522\`):

**1. Test fixture updates.** \`category_selection\` is a required top-level L1 field on this branch (per commit 3fc989d, which flipped it from optional → required to work around OpenAI structured output silently skipping optional fields). Injected the sanctioned zero-placeholder \`{category_id: 0, category_confidence: 0, category_rationale: "not applicable"}\` into ~42 L1 payload literals across 4 pre-existing PR400 test files. No behavior change; purely fixture hygiene. Also updated two \`SYSTEM_PROMPT\` substring assertions in \`test_restart_agent.py\` to reflect that "NVLink" and "Lustre" now appear in the prompt (category 2, 6, 13 name them).

**2. Advisory deferral.** \`repair_invalid_evidence_supports\` is no longer called from \`repair_model_evidence\`. PR #400 already surfaces unknown/duplicate support tags via advisories (\`evidence_support_tag_unknown\`, \`evidence_supports_contain_duplicates\` in \`l1/advisories.py\`) without rejecting the response. Our repair was ported from PR #390 where those advisories didn't exist yet; keeping it enabled would suppress PR #400's advisory mechanism. The repair function itself is retained in \`validation.py\` in case it's needed again.

Full \`tests/attribution/unit/\` suite: **652 passed, 16 subtests passed** with the two logsage-dependent files excluded (they fail identically on stock PR #400 HEAD due to the \`legacy_logsage\` rename; unrelated to this branch).

Ready for review after PR #400 merges to main. If PR #400 rebases again before merge, this branch will need another lightweight rebase — the diff surface with PR #400 is small (mostly \`l1/response_contract.py\`, \`l1/validation.py\`, \`l1/prompts.py\`, and \`l4/policy.py\`).

continue-revolution:
partial review record to cutoff future pr

Separate policy with description
Why id? -- keep it to restrict how llm respond
JSON? -- #419 
Need to consider same issue happen again, sometimes no overwritten (pattern good, but need to consider history as well) -- implemented

id 7 bad token: megatron has the ability to figure out, if megatron cannot figure out, we should stop
id 9 load / async-write complete different, maybe not always restart considering history
id 10 all logs may see this as a secondary, but if valid primary, it should not saying 10
id 15 repeat should not restart

continue-revolution:
Follow-up landed: the L1 category taxonomy has been moved from a Python literal to `l1/categories.json`, loaded and validated at import time via `importlib.resources`. The numeric `id` is retained (LLM structured-output surface, rename-stability, sentinel semantics). 15 new tests cover on-disk shape, invariants, and loader rejection paths. Full attribution/unit suite: 674 passed. PR #419 was closed since the change is now included here as commit `d4a4672`.

continue-revolution:
## Post-reorder eval — confirming no regression

Re-ran the full 79-case × 4-model eval on `gold_v3` after the taxonomy changes (cat 9 narrow, cat 10 harden, cat 13 clarify, cat 36 tighten, cat 37 flip to STOP, drop unused `failure_domain`/`retry_outlook`, and the domain-grouping reorder). Both runs have the opt-in `l1_category_confirmed_restart` two-way context enabled.

### Decision accuracy (the load-bearing metric)

| model | before-reorder | after-reorder | Δ |
|---|---|---|---|
| gpt | 98.7% | **100.0%** | +1.3pp |
| qwen397b | 96.2% | **98.7%** | +2.5pp |
| gemini | 100.0% | 98.7% | −1.3pp |
| nemotron | 92.4% | 92.4% | ±0 |
| **aggregate** | **387.3** | **389.8** | **+2.5pp net** |

All per-model deltas are ≤ 2.5pp, within the LLM run-to-run stochasticity we've observed on every rerun. Net accuracy is up. Gemini's single-case regression is offset by pickups on gpt and qwen; nemotron is flat.

### Category picker accuracy (after only)

Direct picker before/after comparison isn't valid because the pre-reorder eval artifacts stored the *old* category IDs — my scoring uses the new taxonomy's name→ID map, so the pre-reorder artifacts appear to pick "wrong" IDs (they were right against the old numbering). Post-reorder numbers are meaningful in absolute terms:

| model | after-reorder |
|---|---|
| gpt | 94.7% (71/75) |
| qwen397b | 89.5% (68/76) |
| gemini | 90.8% (69/76) |
| nemotron | 80.0% (60/75) |

Absolute picker accuracy is within 2-7pp per model of the pre-reorder run — all consistent with LLM stochasticity, no systematic degradation.

### Policy-context precision

- `l1_category_confirmed_stop` (STOP branch): fires 24/24/24/24 across the four models. Correct: 24/23/23/18. Precision: **100% / 96% / 96% / 75%** (nemotron accumulates its usual residual cluster of progress-log-assertion mispicks).
- `l1_category_confirmed_restart` (RESTART branch, opt-in): fires 0/1/1/0 across the four models. All correct. Precision: **100%**.

### Verdict

Ship. The reorganization is behavior-preserving within LLM noise; on-corpus decision quality is unchanged or slightly better; both policy contexts still fire correctly under the new numbering.

continue-revolution:
## QA review — my colleague's concerns and responses

Category IDs below reference the numbering that was live during our review conversation (pre-reorder). Since then the taxonomy has been reorganized into eight sections and every ID has moved to its new position; see the "Grouping / reorder" section at the end for the mapping.

---

### Cat 7 — Bad token / gradient Inf or NaN
**Concern.** Megatron already has some ability to figure out gradient Inf/NaN issues internally; if it can, we shouldn't see this failure in the first place. When we DO see it, does that mean Megatron already gave up? Should we STOP?

**Response — keep as RESTART.** I dispatched an agent to grep Megatron-LM's source. The load-bearing finding is in `megatron/core/rerun_state_machine.py:534`, inside a branch guarded by `RerunMode.DISABLED`. The default mode is `validate_results`, not `disabled` (`resilience_config.py:21`). In `validate_results`:

1. On the first Inf, Megatron logs (not raises), sets `rerun_requested=True`, and **reruns the same step in place**.
2. If the rerun reproduces Inf → save a checkpoint and `sys.exit(EXIT_CODE_RESUME_TO_DISAMBIGUATE)`.
3. **Megatron itself expects a restart from that checkpoint after this exit code.**

So the `RuntimeError: Unexpected result inf ...` we see in logs is the first-detection raise on the disabled path (or the first attempt on the default path before rerun). It is not a "gave up" signal. RESTART aligns with Megatron's own recovery model.

Recurrence of the same failure across restarts is handled by PR #400's L4 history ledger — the retry budget exhausts and the decision converts to STOP via `exhausted_by`.

---

### Cat 9 — Checkpoint failure / transient checkpoint I/O (load or async-write)
**Concern.** Checkpoint LOAD and checkpoint async-WRITE are functionally different operations. Considering history, a load-error case might not always warrant RESTART.

**Response — narrowed to async-write only.** Empirically the load side was never used: all three gold cases labeled cat 9 in the corpus (b1-042, b1-043, b1-044) are async-write failures (`filesystem_async` inline_container position mismatches, BrokenPipe on write). Renamed to "Checkpoint failure / transient async-write I/O"; description explains that load-side checkpoint I/O errors route to cat 13 (Filesystem I/O) instead — same RESTART decision, cleaner labeling. Zero corpus load-side cases exist, so this is a description tidy-up, not a policy change. Commit `ffab642`.

---

### Cat 10 — Distributed rendezvous / TCPStore socket timeout
**Concern.** TCPStore socket timeout lines appear in many logs as a *secondary* effect (whenever another rank dies, surviving ranks eventually time out on the TCPStore). We shouldn't route to cat 10 in those cascade cases.

**Response — recognized issue, tightened description.** Empirically the picker was already handling cat 10 correctly on the 79-case corpus (10/10 correct picks under the two-way eval), so no observed misclassifications from this pattern. Updated the description to spell out the cascade problem explicitly and instruct the model to pick cat 10 only when TCPStore rendezvous timeout is the *primary observable event* with no earlier crash / OOM / watchdog trigger that would better explain the failure. Preventative for future logs where a cascade IS present. Commit `ed675d8`.

---

### Cat 15 — OSError Errno 98 / port bind race
**Concern.** A repeated port-bind failure across restarts shouldn't keep restarting.

**Response — this is already covered by history.** PR #400's L4 history-based `selected_policy_ledger` counts prior attempts with the same `root_fingerprint` against the retry budget. If the port-bind race recurs without progress, `exhausted_by` fires and `_decision` returns STOP. Our category-driven layer does not need to duplicate that machinery — it's a strict cross-cycle concern that PR #400 already owns.

---

### Cat 12 — Checkpoint failure / corrupt checkpoint metadata on load
**Concern.** The `failure_domain` field on this category is hard to pin down. A corrupted checkpoint file could be workload (bad save code, wrong config); a bit flip during read could be infrastructure. Maybe "unknown" is more honest?

**Response — the `failure_domain` and `retry_outlook` fields have been removed from `CategoryDef` entirely.** After tracing the codebase I confirmed those two fields on every category had zero readers. The L4 policy layer reads `failure_domain` / `retry_outlook_without_workload_change` from the LLM's own `model_recovery_assessment`, NOT from the category taxonomy. So the two taxonomy-level fields were aspirational analyst-reference labels that nothing consumed, and disagreeing with them had no effect on behavior. Dropped them from all 38 entries, `CategoryDef`, the validator, and the JSON schema. Commit `5070311`.

---

### Cat 13 — Filesystem I/O error / Lustre or node-local FS / import visibility
**Concern.** The description mentions "missing generated dataset cache." Is this referring to a missing checkpoint file? But cat 22 is missing-checkpoint-path and goes to STOP — how do these reconcile?

**Response — the confusion was legitimate; rewrote the description.** The two cats are functionally different:

|  | cat 13 (now cat 16) | cat 22 |
|---|---|---|
| **path type** | non-checkpoint (dataset cache, code tree, etc.) | checkpoint path |
| **why "missing"** | transient — filesystem-visibility lag on a network mount (Lustre); file hasn't propagated yet, will be there next attempt | deterministic — config points at a nonexistent checkpoint; retry won't create it |
| **decision** | RESTART | STOP |

Cat 13 groups three non-checkpoint failure surfaces that share the same root cause (network-filesystem visibility on a non-checkpoint path, expected to clear on retry):

1. Lustre I/O stall (`OSError Errno 5` on read/write)
2. Missing generated cache due to filesystem-visibility lag
3. `ImportError` / `ModuleNotFoundError` where the failing import path lives on a network mount — the code tree isn't fully served on that node yet, not a real code bug

Description rewritten to state the non-checkpoint scope up front and explicitly point at cat 22 as the checkpoint-analog. Commit `66ed4cb`.

---

### Cat 24 — CPU OOM / Linux OOM killer
**Concern.** You'd only see the process killed with no reason; there's no signal to distinguish CPU OOM from other causes. Fine to keep, but need to understand.

**Response — the smoking-gun signal exists.** All three corpus cases (b1-001, b1-002, b1-003) contain the SLURM step-log line:

```
slurmstepd ... error: Detected N oom_kill events in StepId=... Some of the step tasks have been OOM Killed.
```

Under two-way eval, all 12 model picks (4 models × 3 cases) correctly landed on cat 24 at 85-96% confidence, explicitly naming `oom_kill` in the mechanism. The ranks themselves only print bare `Killed` lines because SIGKILL by definition leaves no message in the killed process's own output — the "why" lives in the SLURM step log, not in the training-process stdout. The colleague's concern IS real for less-instrumented log captures that drop the slurmstepd output, but our current corpus doesn't exercise that scenario. Description kept concise per your call; the empirical finding is documented in this thread.

---

### Cat 28 — Suspected SDC / DP-replica weight-hash mismatch
**Concern.** What is SDC? Hard to figure out; need to look at the log.

**Response — SDC stands for Silent Data Corruption.** A GPU compute unit (or memory) silently produces wrong bits without triggering any hardware error signal (no Xid, no ECC counter, no watchdog). It's called "silent" because the math just gives you the wrong answer.

Megatron detects it via a periodic check called `check_param_hashes_across_dp_replicas`. It hashes model parameters across DP-replicas; if any rank's hash disagrees with its siblings, it raises an assertion. In our two corpus cases (b1-031, b1-032), the smoking gun is:

```
INFO:megatron.core.utils:[Rank 527] Hash not matching for module.module.decoder.layers.0.mixer.in_proj.weight in model chunk0
INFO:megatron.core.utils:[Rank 527] Hash not matching for module.module.decoder.layers.4.mixer.in_proj.weight in model chunk0
...
[rank391]: assert check_param_hashes_across_dp_replicas(
```

So rank 527's weights for `layers.0.mixer.in_proj.weight` and `layers.4.mixer.in_proj.weight` don't match what its DP peers hold. The most likely explanation is that rank 527's GPU silently corrupted those weights at some point since the last sync — hence "suspected SDC."

But "suspected" is doing work in the name. From the log alone we can't prove SDC vs another divergence source (memory corruption without ECC signal, an optimizer-sync bug, etc.). Decision stays RESTART because restart from the last good checkpoint is plausibly safe if the corruption post-dates it; description kept concise.

---

### Cat 36 — Checkpoint failure / save-path native crash (PyTorchStreamWriter)
**Concern.** Description is much less concise than cat 19. Also: first save should stop, next saves can probably restart. Is that true?

**Response — yes, the first-save-vs-later-save intuition is correct, and it's the load-bearing detail that makes cat 36 STOP-worthy.**

- **First save crashes** → structural. The shape of what we're serializing is fixed by the model config, so the buffer overrun is deterministic against unchanged shape/config. Unchanged retry hits the same crash. **STOP.**
- **Later save crashes** (after at least one successful save) → transient. A prior successful save proves the serializer + shape combination works; a subsequent crash is more likely external (filesystem hiccup, storage congestion, partial-write race). **RESTART.**

This is the same "concrete confirmation" pattern PR #400 uses — a prior successful traversal of a code path is strong evidence the path is sound.

The taxonomy already implements this split — cat 36 is intentionally scoped to the first-save case; later-save transient failures route to cat 9 (async-write I/O) and get RESTART. Tightened the description from 304 → 178 chars, matching cat 19's shape; corpus-specific analyst notes ("iteration 250", "2 samples = 1 unique file idx 107 = idx 381") dropped. Zero corpus cases → no eval risk. Commit `27bdc40`.

---

### Cat 37 — Native heap corruption / glibc malloc abort
**Concern.** Description says pre-first-iter and "definitively software" — sounds like a workload issue, probably a strong STOP case. Why RESTART?

**Response — you're right, flipped to STOP.** The prior state was internally inconsistent: description committed to "definitively software" + "pre-first-iter" but decision was RESTART. The glibc heap-integrity abort (`malloc(): unaligned tcache chunk detected`) fires when malloc's tcache walk finds a corrupted chunk; the corruption is caused by a prior software bug in the process (buffer overflow, double-free, allocator/library ABI mismatch). Combined with the pre-first-iter timing (no successful traversal of the initialization path yet), an in-place retry against unchanged code / library / container image will reproduce the same failure. Flip to STOP. Commit `789702d`.

Long-term improvement note in the commit message: the corpus example was "single-rank" only, hinting at a node-specific cause (corrupt library on that node's local storage, partially updated container image, or bad DRAM). If L4 later adds a "restart with node reallocation" action distinct from "in-place restart," node-specific cases like this could be revisited — they might be safe to RESTART under a fresh allocation while still STOPping under in-place retry.

---

### General — "separate policy from description"
Failure-domain and retry-outlook fields were removed from `CategoryDef` (see cat 12 response). Each category now carries only `id`, `name`, `description`, `decision` — the fields runtime code actually consumes. Commit `5070311`.

---

### General — "group all checkpoint categories together"
Reorganized into eight sections in `categories.json`. All 8 checkpoint categories now sit contiguously in cat 19–26:

```
A. External termination / non-fault      cat 1–6
B. Infrastructure hardware               cat 7–8
C. Container / setup                     cat 9
D. Distributed / networking              cat 10–14
E. Non-checkpoint storage / filesystem   cat 15–18
F. Checkpoint                            cat 19–26  ← contiguous checkpoint block
G. Workload / model                      cat 27–33
H. Native crashes / SDC                  cat 34–38
```

The five descriptions with cross-references to other cats by number were rewritten in place so they point at the new numbers:

- cat 6 → new cat 15 (references old cat 13 → new cat 16)
- cat 9 → new cat 19 (references old cat 4 → new cat 10, and old cat 13 → new cat 16)
- cat 27 → new cat 26 (references old cat 25 → new cat 32)
- cat 13 → new cat 16 (references cat 22, which happens to still be 22)
- cat 16 → new cat 28 (references old cat 13 → new cat 16)

Test constants updated: `RESTART_CATEGORY_ID` 14 → 21, `STOP_CATEGORY_ID` 25 → 32. Loader still validates 1-based sequential IDs. Commit `551604d`.

---

### Post-reorder eval

See the separate comment above for the before/after decision-accuracy scoreboard. TL;DR: aggregate +2.5pp across the 4-model panel with no per-model regression larger than LLM run-to-run noise.

continue-revolution:
the ai slop is way too long. I kinda don't like it. here is a TLDR to address @namitdhameja 's concern.

Id 7 bad token: megatron has the ability to figure out, if megatron cannot figure out, we should stop – keep, because megatron only log, no retry  

Id 9 load / async-write complete different, maybe not always restart considering history – remove load, corpus all async write  

Id 10 all logs may see this as a secondary, but if valid primary, it should not say 10 – no case affected, recognize issue. Updated category description to narrow down pick  

Id 15 repeat should not restart – should have been addressed by history  

Id 12 failure domain probably unknown because ckpt bad \-\> workload issue, bit flip \-\> infra, first call, infrastructure probably better – failure domain label not affect decision at all, remove these labels for now  

Id 13 need to read the log to know what it is, is it referring to checkpoint missing file? But id 22 is referring to it and it goes to stop. –

|  | cat 13 | cat 22 |
| :---- | ----- | ----- |
| **path type** | non-checkpoint (dataset cache, code tree, etc.) | checkpoint path |
| **why it "went missing"** | transient — filesystem visibility lag on a network mount (Lustre); file hasn't propagated yet, will be there next attempt | deterministic — config points at a nonexistent checkpoint; retry won't create it |
| **decision** | RESTART | STOP |

  Cat 13 groups three scoped-to-non-checkpoint failure surfaces:  
  1\. Lustre I/O stall (OSError Errno 5 on read/write)  
  2\. Missing generated cache due to filesystem-visibility lag (transient)  
  3\. ImportError / ModuleNotFoundError where the failing import path lives on a network mount — the code tree isn't fully served on that node yet, not a real code bug  
    
  All three share the same root: network-filesystem visibility on a non-checkpoint path, expected to clear on retry.  

Id 24: You won’t see it, process will just be killed, fine to keep – a message is available when this happens: 'slurmstepd ... error: Detected N oom\_kill events in StepId=... Some of the step tasks have been OOM Killed.'  

Id 28: What is it? Hard to figure out SDC issue. Need to take a look at the log file –   
Megatron has a periodic check called check\_param\_hashes\_across\_dp\_replicas. It hashes model parameters across DP-replicas; if any rank's hash disagrees with its siblings, it raises an assertion. In our two corpus  
  cases (b1-031, b1-032), the smoking gun is:

  INFO:megatron.core.utils:\[Rank 527\] Hash not matching for module.module.decoder.layers.0.mixer.in\_proj.weight in model chunk0  
  INFO:megatron.core.utils:\[Rank 527\] Hash not matching for module.module.decoder.layers.4.mixer.in\_proj.weight in model chunk0  
  ...  
  \[rank391\]: assert check\_param\_hashes\_across\_dp\_replicas(

  So rank 527's weights for layers.0.mixer.in\_proj.weight and layers.4.mixer.in\_proj.weight don't match what its DP peers hold. The most likely explanation is that rank 527's GPU silently corrupted those weights at  
  some point since the last sync — that's the "suspected SDC" verdict.

  But "suspected" is doing work in the name. From the log alone we can't prove it was SDC vs some other divergence source (memory corruption without ECC signal, an optimizer-sync bug, etc.).  

Id 36 maybe can be more general (for example, id 19 is good): first save should stop, next all save could probably restart – done  

Id 37: Why restart? Also pre first iter, sounds like a workload issue, probably strong stop case – flipped the decision to STOP  

Separate policy with description – failure domain label removed  

Probably put all checkpoint in one place –  
  \- 5 descriptions with cross-references to other cats need their numbers updated:  
    \- Cat 6 → new 15 (references cat 13 → new 16\)  
    \- Cat 9 → new 19 (references cat 4 → new 10, cat 13 → new 16\)  
    \- Cat 10 → new 12 (references cats 3/38/25 — all shift)  
    \- Cat 13 → new 16 (references cat 22 → new 22, same\!)  
    \- Cat 16 → new 28 (references cat 13 → new 16\)  
Final new layout  
  A. External termination / non-fault      cat 1–6  
  B. Infrastructure hardware               cat 7–8  
  C. Container / setup                     cat 9  
  D. Distributed / networking              cat 10–14  
  E. Non-checkpoint storage / filesystem   cat 15–18  
  F. Checkpoint                            cat 19–26  ← contiguous checkpoint block  
  G. Workload / model                      cat 27–33  
  H. Native crashes / SDC                  cat 34–38

Maybe drop all restart in json we only need to classify stop – do not drop because of explainability and ability to fall back to policy driven decision

docs/design/attribution/restart_agent/spike_l1_category_integration.md
namitdhameja, line 1: I think it fine to have a feature doc, but then would remove branch details from it

src/nvidia_resiliency_ext/attribution/restart_agent/l1/categories.json
```diff
@@ -0,0 +1,233 @@
...
+    },
+    {
+      "id": 19,
+      "name": "Checkpoint failure / transient async-write I/O",
+      "description": "Transient I/O failure during the checkpoint async-save path on a checkpoint directory: filesystem_async 'Local process 0 encountered an error', PyTorch inline_container position mismatch, BrokenPipeError Errno 108 from a writer, OSError Errno 5 (Input/output error). Prefer this over cat 16 whenever the failing path is a checkpoint directory and the mechanism is on the write side, even if the surface is generic filesystem I/O. Load-side checkpoint I/O errors that do not fit cat 10 (early/startup timeout) route to cat 16 (Filesystem I/O) instead; they have the same decision (RESTART) but a different labeling story.",
+      "decision": "RESTART"
+    },
+    {
+      "id": 20,
+      "name": "Checkpoint failure / corrupt checkpoint metadata on load",
```
namitdhameja, line 120: This is fine, but make sure to test with an actual corrupt checkpoint, where even if the 1st call is restart; but on the repeat it detects as a persistent failure and stops

src/nvidia_resiliency_ext/attribution/restart_agent/l1/categories.json
```diff
@@ -0,0 +1,233 @@
...
+    },
+    {
+      "id": 30,
+      "name": "CUDA illegal op in CUDA graph capture",
+      "description": "Illegal CUDA operation during graph capture; deterministic code bug on the same code path.",
+      "decision": "STOP"
+    },
+    {
+      "id": 31,
+      "name": "CPU OOM / Linux OOM killer",
```
namitdhameja, line 186: As discussed, linux OOM is not in the app log. I think it's in the slurm output. Its fine to have this here but would likely not get activated

src/nvidia_resiliency_ext/attribution/restart_agent/l1/categories.py
```diff
@@ -0,0 +1,110 @@
...
+
+from __future__ import annotations
+
+import json
+from dataclasses import dataclass
+from importlib import resources
+from typing import Any
+
+CATEGORIES_SCHEMA_VERSION = "restart_agent_categories.v1"
+_ALLOWED_DECISIONS = frozenset({"STOP", "RESTART", "EXCLUDED"})
```
namitdhameja, line 24: Is excluded referring to node exclusion?

src/nvidia_resiliency_ext/attribution/restart_agent/l1/validation.py
```diff
@@ -317,3 +325,157 @@ def _positive_line_errors(value: Any, field: str) -> list[str]:
...
 def _nonempty_string(value: Any) -> bool:
     return isinstance(value, str) and bool(value.strip())
+
+
+def _category_selection_errors(value: Any) -> list[str]:
+    """Validate the optional L1 category_selection block.
+
+    The block is optional. When present, it must have all three curated fields
+    (category_id, category_confidence, category_rationale). category_id must be
+    an integer in the range [0, 38] (0 means "no listed category matches").
```
namitdhameja, line 335: Maybe ok to just say range instead of 0, 38; only saying as the range may change esp as now is loaded from a file and not inherent in the code

src/nvidia_resiliency_ext/attribution/restart_agent/l1/validation.py
```diff
@@ -317,3 +325,157 @@ def _positive_line_errors(value: Any, field: str) -> list[str]:
...
+            f"schema_version: normalized {actual!r} -> {L1_EVIDENCE_SCHEMA_VERSION!r} "
+            f"(model hallucinated similar version string)"
+        )
+    return notes
+
+
+def repair_biconditional_unknowns(payload: Any) -> list[str]:
+    """In-place fix: normalize model_recovery_assessment claim (value, status) pairs.
+
+    Contract rule: value == "unknown" iff status == "unknown". Some models
```
namitdhameja, line 405: iff :)

src/nvidia_resiliency_ext/attribution/restart_agent/l4/policy.py
```diff
@@ -569,11 +582,152 @@ def _effective_policy(
...
+
+
+def _match_l1_category_context(
+    *,
+    primary: FailureEvidence | None,
+    l1_category_selection: Mapping[str, Any] | None,
+) -> _PolicyContextMatch | None:
+    """Match the L1 category taxonomy as a policy_context.
+
+    The LLM classifies into the 38-entry taxonomy at l1/categories.py; when it
```
namitdhameja, line 595: Its a repeat comment, but scan through and remove 38

src/nvidia_resiliency_ext/attribution/restart_agent/l4/policy.py
```diff
@@ -569,11 +582,152 @@ def _effective_policy(
     )
 
 
+L1_CATEGORY_CONFIRMED_STOP_CONTEXT_ID = "l1_category_confirmed_stop"
```
namitdhameja, line 585: The meta comment is that to make sure to test that the right think happens on sticky errors ie
Bad checksum file gets a retry, but on 2nd invocation of same moves to a STOP; early nccl timeout gets a restart but would STOP after <num> retries, same for bad token.

namitdhameja (changes requested):
good show - small nits !

src/nvidia_resiliency_ext/attribution/restart_agent/l1/categories.json
continue-revolution: 'slurmstepd ... error: Detected N oom_kill events in StepId=... Some of the step tasks have been OOM Killed.'

this is typically used to detect linux oom. it's seen in qa provided logs. would we see this message in a realistic log?

namitdhameja: That depends on whether they are feeding the app cycle log or there was a run with no nvrx so they used the complete run slurm log. The potential AI is to consider Slurm log, but thats not tied to this commit.

src/nvidia_resiliency_ext/attribution/restart_agent/l1/categories.py
continue-revolution: no, it means the category is excluded from the eval-metric / policy-activation surface (e.g. no failure observed, incomplete/truncated log, or a multi-cycle NVRx-managed run).

I think node exclusion is a good decision to add later upon some exploration

src/nvidia_resiliency_ext/attribution/restart_agent/l4/policy.py
continue-revolution: same reason restart has been made sure to stop at design level, see [here](https://github.com/NVIDIA/nvidia-resiliency-ext/pull/411/changes#diff-b9bcb577f5cfc990fd45bb6405e93892e7b475b6588be1f3e4e0a2a31625d0f9R665)

namitdhameja (approved):

PR #415, opened 2026-09-01
Confine gRPC log server writes to allowed roots

hexinw-nvidia:
## Problem

`LogAggregationServicer.StreamLogs` passed the client-supplied `LogChunk.file_path`
straight to `os.makedirs` + `open(path, 'ab')` with no confinement — no allowed root, no
rejection of absolute paths or `..`, no symlink check. `ft_launcher` auto-spawns this
server bound to `0.0.0.0` via `add_insecure_port`, so any peer that could reach the funnel
port could create directories and append attacker-controlled bytes to any file writable by
the training process. The leaf server forwarded the same client-controlled path to the root.

## Fix

- `resolve_under_allowed_roots()` in `shared_utils/os_utils.py` — `realpath` + `commonpath`
  confinement. Returns the resolved path, and that is what gets opened, so a symlink planted
  between check and open cannot redirect the write.
- `allowed_roots` is **keyword-only with no default** on both servicers, so there is no
  unconfined mode to fall into. Rejected chunks abort the stream with `INVALID_ARGUMENT`.
- Both CLIs take a repeatable, required `--allowed-root`. The launcher derives the roots and
  passes them to the root and every leaf; if they cannot be resolved it fails closed, falling
  back to direct file writing rather than running unconfined.

## Scope decisions

Roots are `--ft-per-cycle-applog-prefix` and `--ft-nvrx-logfile` only — the two destinations
clients actually target over gRPC. Deliberately excluded:

- `--ft-log-server-log-prefix` — `_root.log` / `_leaf_N.log` are written by the launcher
  redirecting each subprocess's stdout/stderr, never through a `LogChunk`.
- `--ft-cycle-info-dir` — `CycleInfoWriter` writes its JSON and symlink directly.

Including either would widen the boundary for no functional gain.

Root directories are created before being resolved: `realpath()` of a missing directory
returns it unchanged, so a directory that only later appeared as a symlink would resolve
elsewhere at write time and the job's **own** logs would be rejected.

## Performance

No steady-state cost. The root validates inside the branch that already does the
`makedirs`/`open`, so it runs once per distinct path — chunk handling stays a single dict
lookup. The leaf memoizes per RPC and evicts FIFO, because the newest path (the current
cycle log) is the hot one.

## Testing

- New `tests/shared_utils/test_os_utils.py`; confinement coverage added to the existing
  `test_grpc_log_server.py`, `test_grpc_log_leaf_server.py`, and `test_launcher.py` so the
  layout keeps mirroring modules.
- The reported reproducer plus variants (`..` traversal, symlinked dir, symlinked file,
  `~/.ssh/authorized_keys`, `/etc/cron.d`, sibling-prefix root, relative path, NUL byte,
  empty path) verified blocked against both servers; legitimate writes unaffected.
- No regressions against the pre-change baseline; `black` / `isort` / `ruff` clean.

## Note for reviewers

Per CONTRIBUTING this needs a tracked issue and the `#<Issue Number> - <Title>` commit
prefix — please supply the issue number and I will amend. Two pre-existing items are
**not** addressed here and may deserve their own issues: the log client has no backoff on
`grpc.RpcError`, and in two-level mode the root binds `0.0.0.0` although only same-host
leaves connect to it.

### DIFF
--- a/src/nvidia_resiliency_ext/fault_tolerance/control_plane.py
+++ b/src/nvidia_resiliency_ext/fault_tolerance/control_plane.py
@@ -161,7 +161,24 @@ def _start_control_services(
         applog_prefix = args.ft_per_cycle_applog_prefix
         log_funnel_ports = None
         grpc_log_prefix = None
-        if args.ft_enable_log_server:
+        # The funnel writes to paths its clients choose, so it can only run confined to the
+        # directories those logs live in. --ft-log-server-log-prefix names where the funnel's
+        # own diagnostics go and cannot define that boundary. Co-located control gets this
+        # structurally (the launcher only funnels inside `if base_log_file:`); mirror it here
+        # rather than failing startup, since log funneling is on by default for nvrx-control.
+        log_funnel_enabled = bool(args.ft_enable_log_server) and bool(
+            args.ft_per_cycle_applog_prefix or args.ft_nvrx_logfile
+        )
+        if args.ft_enable_log_server and not log_funnel_enabled:
+            logger.warning(
+                "Log funneling is enabled but neither --ft-per-cycle-applog-prefix nor "
+                "--ft-nvrx-logfile is set, so there is no client log destination to confine "
+                "the funnel to; not starting the gRPC log server(s). Launchers that are "
+                "themselves configured to funnel will block waiting for this host's funnel "
+                "port -- they do not fall back to direct writes. Pass the same applog prefix "
+                "here, or disable --ft-enable-log-server on those launchers."
+            )
+        if log_funnel_enabled:
             grpc_log_prefix = _resolve_grpc_log_server_log_prefix(args)
             assert grpc_log_prefix is not None
             log_funnel_ports = LogFunnelPorts.from_launcher_args(args)
@@ -182,7 +199,7 @@ def _start_control_services(
                 )
                 services.attribution_service.start_poller()
 
-        if args.ft_enable_log_server:
+        if log_funnel_enabled:
             assert log_funnel_ports is not None
             assert grpc_log_prefix is not None
             services.grpc_processes = _start_grpc_log_servers(
--- a/src/nvidia_resiliency_ext/fault_tolerance/launcher.py
+++ b/src/nvidia_resiliency_ext/fault_tolerance/launcher.py
@@ -3255,6 +3255,48 @@ def _grpc_log_path(log_prefix: str, suffix: str) -> str:
     return log_prefix + suffix
 
 
+def _resolve_grpc_log_allowed_roots(args: Any) -> List[str]:
+    """
+    Directories the gRPC log servers are allowed to write into.
+
+    The servers bind an unauthenticated port and write to a path chosen by each client, so
+    they are given exactly the directories that gRPC clients actually target: the per-cycle
+    application log prefix and the launcher log. Per-cycle files (``<prefix>_cycleN.log``)
+    stay in the prefix's directory, so the dirname covers every cycle.
+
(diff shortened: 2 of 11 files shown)


hexinw-nvidia:
> **P1 — Accepted configuration breaks startup** (`launcher.py:3432`)

Valid finding, fixed in 899d4a1. Thanks — this caught a real gap: I had only traced the `launcher.py:3122` caller of `_start_grpc_log_servers`, where the funnel is already nested inside `if base_log_file:` so an applog prefix is guaranteed. `control_plane.py:188` is a second caller with no such guard, and it turns an empty result into `RuntimeError`.

One correction to the suggested direction: restoring `--ft-log-server-log-prefix` as an allowed root would **not** fix this. In that configuration the clients are launchers writing to *their own* applog prefix, which the control node doesn't know — confining to the control node's own diagnostics directory would start cleanly and then reject every client chunk. Silent total log loss is worse than the loud failure.

Rejecting during argument validation also isn't right here: `add_log_funnel_args(parser, enable_default=True)` means funneling is **on by default** for `nvrx-control`, so requiring an applog prefix would force it on cycle-info-only deployments (see `test_control_parser_accepts_cycle_info_dir_with_job_id_without_applog_prefix`).

So the fix mirrors what co-located control already does structurally — run the funnel only when there is a client log destination to confine it to, and warn when skipping:

```
Log funneling is enabled but neither --ft-per-cycle-applog-prefix nor
--ft-nvrx-logfile is set, so there is no client log destination to confine
the funnel to; not starting the gRPC log server(s). Clients will write to
Lustre directly.
```

Validation is unchanged, so no previously accepted configuration is rejected. Added `test_control_skips_log_funnel_without_confinable_destination`; `nvrx-control` now starts cleanly in the reported configuration with the funnel skipped rather than raising.

hexinw-nvidia:
> **P1 — Disabled funnel strands remote logs** (`control_plane.py:171`)

The mechanism is real but I don't think this is a legitimate configuration, so I'm not changing the gate. I did fix the misleading warning it exposed (5a27dd3).

**Confirmed mechanism:** `_wait_for_server_ready` retries forever on `grpc.RpcError` (no max attempts, no timeout), and the "falling back to direct file writing" path at `launcher.py:3126-3132` sits inside `if is_tcp_store_host:`, so remote launchers have no fallback. Accurate.

**Why it isn't a supported deployment:**

1. It's a config mismatch. The control node is the process that *writes* the aggregated file, so it must know where that file goes. Enabling funneling on launchers while denying control the destination is an operator error, and the same operator sets both halves.
2. Making it work again means reverting this PR. That configuration only functioned before because the control node accepted whatever path clients sent — that is the vulnerability being fixed. There is no version where control writes client logs correctly without knowing the destination.
3. The stranding is pre-existing and general, not introduced here. Remote launchers hang identically today if the root server crashes at startup, the funnel port is firewalled, or `nvrx-control` isn't running at all. The gate is one more route into an already-reachable state.

Per the maintainers, production log funneling is done either through co-located control or external control; a control node that funnels for launchers whose log destination it doesn't share is not a deployment mode.

**What I did take from this:** the warning said *"Clients will write to Lustre directly"*, which is true only for launchers that aren't funneling — in exactly the case you describe they block instead. Corrected:

```
... not starting the gRPC log server(s). Launchers that are themselves
configured to funnel will block waiting for this host's funnel port -- they
do not fall back to direct writes. Pass the same applog prefix here, or
disable --ft-enable-log-server on those launchers.
```

The absent client-side timeout/fallback is worth its own issue, independent of this PR.

sbak5 (approved):

PR #419, opened 2026-09-03
L1 category taxonomy: move from Python literal to JSON

continue-revolution:
## Summary

Extracts the 38 L1 category definitions from an inline Python literal into a JSON file colocated with the module. The Python module now loads and validates the JSON at import time via `importlib.resources`.

**Behavior is unchanged.** `CATEGORIES` still exposes the same immutable tuple of `CategoryDef`; `category_by_id` still returns the same rows; the numeric `id` field is preserved.

## Why

Reviewer feedback: the taxonomy is data, not code, and belongs in a data file. Benefits:

- Non-engineers can extend the taxonomy without touching Python.
- Diffs are easier to review and cross-reference against gold labels or other tools.
- Other tools (offline analyzers, dashboards, docs generation) can reuse the JSON directly.

## Why the numeric `id` field stays

I kept it. The functional reasons:

1. **LLM structured-output surface.** The category picker asks the LLM to emit `category_id: integer` in `[0, 38]` (see `response_contract.py`). A bounded integer is materially more robust than an exact free-text name for structured emission — models drop articles, reformat punctuation, and truncate long strings, and the schema range check catches all of that. Enums of 38 exact strings work in structured output too but are noisier in practice.
2. **Stable identifier across renames.** If a category is reworded later, historical eval artifacts that stored `category_id=25` still point at the correct row. If we keyed on name, a rename would break the historical link or require an alias table.
3. **`category_id=0` is the sanctioned \"no listed category matches\" sentinel.** Used ~20 times in test fixtures and documented in the response contract. Replacing it would require plumbing a new sentinel through the schema and every fixture.

If a reviewer disagrees and wants to drop the id anyway, we can do that in a follow-up — it's a substantially larger diff.

## What's in the PR

- `l1/categories.json` — 38 entries with fields `{id, name, description, decision, failure_domain, retry_outlook}` and a `schema_version` string.
- `l1/categories.py` — reduced from 493 lines to ~120. Loads via `importlib.resources.files(__package__).joinpath('categories.json')`. Validates: schema_version, required fields, id is a positive integer, ids are 1-based sequential, names are unique, decision ∈ `{STOP, RESTART, EXCLUDED}`, all string fields are strings. Raises `ValueError` on any violation so a bad taxonomy fails at import, not later.
- `tests/attribution/unit/test_restart_agent_l1_categories_json.py` — 15 tests covering on-disk shape, ID/name/decision invariants, `category_by_id` semantics, and every loader-rejection branch.

## Base branch

Based on `chengsongz/decision-from-category-on-pr400` (PR #411), which introduces the taxonomy. Once PR #411 merges, this PR will be retargeted to `main`.

## Test plan
- [x] `pytest tests/attribution/unit/` — 674 passed on my machine
- [x] `black --check`, `isort --check-only`, `ruff check` on touched files
- [x] Verified the JSON ships in the installed wheel via `importlib.resources`
- [x] Smoke-verified `CATEGORIES`, `category_by_id`, and `CATEGORIES_SCHEMA_VERSION` are still importable from the same paths

### DIFF
--- a/src/nvidia_resiliency_ext/attribution/restart_agent/l1/categories.json
+++ b/src/nvidia_resiliency_ext/attribution/restart_agent/l1/categories.json
@@ -1,0 +1,309 @@
+{
+  "schema_version": "restart_agent_categories.v1",
+  "categories": [
+    {
+      "id": 1,
+      "name": "SLURM step cancelled / drain / preemption",
+      "description": "Clean external cancellation by the scheduler (scancel, drain, preemption); no application fault.",
+      "decision": "RESTART",
+      "failure_domain": "infrastructure",
+      "retry_outlook": "may_recover"
+    },
+    {
+      "id": 2,
+      "name": "NVLink / GPU error",
+      "description": "Uncorrectable NVLink/GPU hardware error (Xid, ECC DBE, peer-access failure); the device fault may be transient and clear on an in-place restart.",
+      "decision": "RESTART",
+      "failure_domain": "infrastructure",
+      "retry_outlook": "may_recover"
+    },
+    {
+      "id": 3,
+      "name": "Node failure / SLURM externally terminated",
+      "description": "SLURM 'CANCELLED DUE TO NODE FAILURE' or other external node termination.",
+      "decision": "RESTART",
+      "failure_domain": "infrastructure",
+      "retry_outlook": "may_recover"
+    },
+    {
+      "id": 4,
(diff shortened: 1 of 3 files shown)


continue-revolution:
Merged into PR #411 via commit `d4a4672` (Move L1 category taxonomy to JSON). The taxonomy-JSON conversion is now part of PR #411's tree. Closing this PR as the change lands with PR #411.

PR #422, opened 2026-09-09
Add v0.7 release notes and migration guidance

hexinw-nvidia:
## Summary

- add curated v0.7.0 release notes covering shipped FT, attribution, health-check, checkpointing, logging, packaging, and security changes
- restore the v0.6.0 release-note section that was finalized only on release/v0.6 and never carried into main
- document that barrier rendezvous is the only supported FT rendezvous implementation and provide migration guidance for removed legacy mode
- correct the singleton deployment entry page: distinguish container and bare-metal prerequisites, document the ClusterUUID requirement and simple-mode opt-out, describe array hand-off and shape accurately, and name the watcher action

## Release workflow

This PR targets main first so the cumulative release history and user guidance carry into future release branches. After merge, the focused documentation commit will be backported to release/v0.7.

## Validation

- verified every PR referenced by the v0.7 notes is present in release/v0.7
- verified the restored v0.6 section matches the v0.6.0 release version
- git diff --check
- poetry check --lock
- Sphinx HTML build completed successfully; the repository still reports pre-existing documentation warnings, with none against the updated singleton page or added usage-guide section


docs/source/release-notes.md
```diff
@@ -2,6 +2,147 @@
...
+    - Attribution analysis is asynchronous and does not delay restart. Acting on a `STOP` verdict is opt-in through `--ft-attribution-stop-action no-restart`; the default `log` mode records verdicts without terminating the job ([#378](https://github.com/NVIDIA/nvidia-resiliency-ext/pull/378)).
+    - The new **Restart Agent** provides deterministic failure evidence and retry policy, with optional model-assisted enrichment, bounded history, operational logging, and auditable result artifacts ([#370](https://github.com/NVIDIA/nvidia-resiliency-ext/pull/370), [#400](https://github.com/NVIDIA/nvidia-resiliency-ext/pull/400)).
+    - The core wheel includes `attrsvc`, Flight Recorder analysis, and the direct Restart Agent backend. MCP, `setproctitle`, and Slack integrations require `nvidia-resiliency-ext[attribution]`; legacy LogSage remains source-checkout-only and is excluded from wheels ([#404](https://github.com/NVIDIA/nvidia-resiliency-ext/pull/404), [#410](https://github.com/NVIDIA/nvidia-resiliency-ext/pull/410)).
```
chenming1986, line 22: **setproctitle** seems like an implementation detail rather than a user-facing feature

hexinw-nvidia: Agreed. `setproctitle` only labels the MCP analysis-server process and is not a user-facing capability. Removed it from the release-note summary in 66e3bfe while retaining the MCP/Slack optional-extra boundary.

PR #424, opened 2026-09-14
segment-health: serve() loop with pidfile + SIGUSR1 immediate-poll

hexinw-nvidia:
## What

Adds a `scheduler_segment_health_serve()` entry point to the SLURM segment-health producer (`examples/fault_tolerance/deployment/slurm/scheduler_segment_health.sh`) that centralizes the poll cadence and gives operators a fast-path.

## Why

When an operator drains a suspect node, the drain should reach the per-task control files in **seconds, not up to one poll interval** (default 300s). Today each caller reimplements the `configure` + `poll_once` + `sleep` loop, and there is no way to force an early poll.

## How

`serve()`:
- runs the long-lived poll loop (`poll_once` every `interval`, default `SEGMENT_HEALTH_POLL_INTERVAL` / 300s);
- publishes `host + PID` to `${NVRX_SEGMENT_HEALTH_CHECK_DIR}/segment_health_producer.<array_job_id>.pid` (uses `$BASHPID` since it is meant to run backgrounded; host recorded because the poller runs on array task 0's node, which the operator reaches over ssh);
- traps **SIGUSR1** to interrupt the sleep and force an immediate `poll_once` — `poll_once` is idempotent, so an on-demand poll is safe and is a no-op when Slurm state is unchanged;
- traps TERM/INT/EXIT for best-effort teardown (removes pidfile + temp workspace).

Interruptible sleep is implemented by backgrounding `sleep` and `wait`-ing on it so SIGUSR1 returns the wait early.

## Compatibility

Additive and backward compatible: `configure` / `poll_once` / `cleanup` are unchanged, so existing inline-loop callers keep working. This only factors the cadence, signal handler, and pidfile into one place so callers need not reimplement them.


hexinw-nvidia:
Ready for review. All four Greptile findings are addressed in `bbf4e31`:

- **P1 — SIGUSR1 lost:** USR1 handler now sets a `poll_pending` flag; the loop re-polls immediately instead of sleeping, so a signal arriving during `poll_once` is never dropped.
- **P2 — pidfile ownership:** teardown removes the pidfile only when it still records this `$BASHPID` (a successor producer that republished it is not clobbered).
- **P2 — invalid interval leaked workspace:** interval is validated *before* `configure` allocates the temp workspace.
- **P2 — trap quoting:** teardown moved to a function that reads `$pidfile` at trap time (nothing interpolated into the trap string; apostrophe-safe).

Validated end-to-end on CMH (GB200 NVL72), 2‑rack + 3‑segment hot-spare InJob runs:
- **L1** — serve() selected via `declare -F`; pidfile created and cleanly removed on teardown.
- **L2** — SIGUSR1 → immediate poll (tick landed well inside the 300s interval).
- **L3** — operator drain → exclusion → `ft_launcher` refuses the excluded segment (`Segment health check failed … excluded by scheduler state` → `UnhealthyNodeException`) → standby spare promoted, world size held.

`bash -n` clean. Backward compatible: `configure`/`poll_once`/`cleanup` unchanged, so existing inline-loop callers are unaffected.

namitdhameja (approved):

PR #427, opened 2026-09-16
Add Slack notifications for attribution results in nvrx-smonsvc

continue-revolution:
## Problem

A deployed restart-agent stack cannot alert anyone. There is no Slack path on the direct Restart Agent backend:

- `attribution/restart_agent/` contains no Slack code at all — it never notifies.
- `nvrx-attrsvc`'s `lib` backend does not invoke the postprocessing pipeline that owns the existing Slack integration.
- Those hooks (`attribution/postprocessing/slack.py`) are reachable only from `legacy_logsage`'s `AttributionController`, which is excluded from built wheels and is no longer a selectable attrsvc backend.
- `SLACK_BOT_TOKEN` / `SLACK_CHANNEL` still exist in attrsvc `Settings` but nothing on this path reads them — setting them is a silent no-op.

Meanwhile `nvrx-smonsvc` already fetches a normalized `recommendation` for every terminal job and does nothing with it but print to stdout.

## Change

Notify from `nvrx-smonsvc`, where the data already is.

- **New** `services/smonsvc/slack.py`: `SlackConfig`, `SlackNotifier`, and the record adapter. Reuses `format_posting_markdown_body` for the message body and `get_slack_user_id` for owner mentions, so alerts match the format the posting pipeline produces.
- Thread an optional notifier through `fetch_results` → `log_attribution_result`; construct it in `SlurmJobMonitor`.
- Expose delivery counters under `slack` in `/stats`.
- Drive-by: add the missing newline before the `*Log path:*` section in `posting_markdown.py`, which ran into the preceding code block. Affected the legacy path too.

### Message shape

```
*NVRx attribution:* `STOP` _(source: log_analyzer)_
*Reason:* uncorrectable ECC error on gpu 3, node h100-0421
*Job ID:* `5728163 (nemotron5_340b_pretrain)`
*Failed due to:*
...
*Terminal issue:*
...
*Log path:*
...
```

## Configuration

| Variable | Default | Description |
|---|---|---|
| `SLACK_BOT_TOKEN` / `SLACK_BOT_TOKEN_FILE` | `""` | Bot token; falls back to `~/.slack_bot_token`, `~/.slack_token`, `~/.config/nvrx/slack_bot_token` |
| `SLACK_CHANNEL` | `""` | Channel ID or name |
| `NVRX_SMONSVC_SLACK_NOTIFY_ACTIONS` | `STOP` | Actions that page |

Requires `slack-sdk`, already declared in the `attribution` extra.

## Behavior notes

- **Off by default.** Activates only when `slack-sdk` is installed *and* both a token and channel are set; otherwise the monitor logs `Slack alerts: disabled (...)` at startup and behaves exactly as before.
- **Only `STOP` pages by default**, matching the existing `should_notify_slack`. `RESTART` is the routine outcome and would be noisy.
- **Unrecognized action names are dropped with a warning** rather than passed through `normalize_recommendation_action`, which maps unknown input to `UNKNOWN` and would otherwise page on every unattributed job.
- **Delivery is best effort.** Notifier errors are counted and logged, never propagated into the polling loop — a Slack outage must not stop the monitor from marking results fetched.

## Docs

smonsvc README gains a Slack section. The attrsvc README is corrected — it asserted Slack was unavailable on this path. Env vars are documented in `run_services.sh`, `nvrx_services.sbatch`, `scripts/README.md`, and the `setup_systemd.sh` `nvrx.env` template (systemd units do not inherit the shell environment, so that one matters for real deployments).

## Testing

- 23 new tests in `services/tests/test_smonsvc_slack.py` covering config parsing, record/message shape, notify gating, failure handling, and integration with `log_attribution_result`.
- 791 passed across `services/tests/` and `tests/attribution/unit/`.
- `black --check`, `isort --check-only`, `ruff check` clean repo-wide.
- Smoke-tested the real path with `slack-sdk` installed: token/channel resolution, action parsing, and `WebClient` construction.

Two pre-existing failures in my environment, unrelated and confirmed by stashing: `test_prepare_uds_path_permission_error_does_not_unlink` (sandbox errno differs) and a missing `fastapi` until installed locally.


services/attrsvc/README.md
```diff
@@ -76,16 +76,47 @@ Restart Agent backend.
...
-| `SLACK_CHANNEL` | `""` | Channel ID or name (e.g. `#trng-alerts`). In `.env`, quote values that start with `#`: `SLACK_CHANNEL="#trng-alerts"` |
+| `SLACK_BOT_TOKEN` | `""` | Bot token. If empty, falls back to `SLACK_BOT_TOKEN_FILE`, then `~/.slack_bot_token`, `~/.slack_token`, `~/.config/nvrx/slack_bot_token` |
+| `SLACK_BOT_TOKEN_FILE` | — | Path to a file containing the token |
+| `SLACK_CHANNEL` | `""` | Channel ID or name (e.g. `#trng-alerts`). In `.env` files, quote values starting with `#` |
+| `NVRX_ATTRSVC_SLACK_NOTIFY_ACTIONS` | `STOP` | Comma- or space-separated actions that trigger a message. Valid: `STOP`, `RESTART`, `CONTINUE`, `UNKNOWN`, `TIMEOUT` |
+
+Requires `slack-sdk`:
+
+```bash
+pip install 'nvidia-resiliency-ext[attribution]'
```
namitdhameja, line 103: You could probably just have the slack sdk in mainline nvrx? The attribution bring in mcp et all as well - which folks don't like as has conflicts

services/attrsvc/README.md
```diff
@@ -76,16 +76,47 @@ Restart Agent backend.
...
+Notifications are **off by default**, activating only when `slack-sdk` is
+installed *and* both a token and channel are configured; otherwise attrsvc logs
+`Slack alerts: disabled (...)` at startup and behaves unchanged. Delivery is best
+effort — a Slack outage is logged, never propagated into the analysis path.
+
+Each message carries the recommendation and its source, the job ID, the narrative
+root cause, the L4 decision trail (rule, category, retry outlook), an evidence
+line locating the failure, and — when the cause is unconfirmed — the plausible
+causes and missing evidence. The job owner is mentioned when their
+`{user}@nvidia.com` address resolves to a Slack account.
```
namitdhameja, line 115: for internal @nvidia is fine, but you don't want to presume that as this is a public lib and is valid scenario to run for non-nv user as well

services/scripts/README.md
```diff
@@ -52,6 +52,39 @@ The default direct backend accepts `LLM_API_KEY_FILE`, an explicit Restart
...
+Under systemd, set these in `nvrx.env` instead — units do not inherit your
+shell environment.
+
+### Application Log Resolution
+
+If SLURM `StdOut` points at a launcher wrapper rather than the training log,
+enable resolution so the monitor submits the real log:
+
+```bash
+export NVRX_SMONSVC_APP_LOG_RESOLUTION=1
```
namitdhameja, line 80: attrsvc has SPLITLOG mode to detect the case where the .out and .log files are different. Thats not with smonsvc though. Not sure if are using same so perhaps thats why

services/attrsvc/README.md
continue-revolution: ok, now slack-sdk is a regular dependency.

services/attrsvc/README.md
continue-revolution: fixed

services/scripts/README.md
continue-revolution: looks like this SPLITLOG is a logsage thing and our new restart agent never use this variable. also nvrx inline already select the proper log path through [ft_rendezvous_barrier](https://github.com/NVIDIA/nvidia-resiliency-ext/blob/main/src/nvidia_resiliency_ext/fault_tolerance/ft_rendezvous_barrier.py#L1688) and submit to attrsvc, so this `NVRX_SMONSVC_APP_LOG_RESOLUTION` variable is only for smonsvc and `log_resolver` which is to figure out where `.log` only belongs to smonsvc.

